PR #32, opened 2022-04-09
Fixed issues with duplicate `IServiceProvider` registration

Aaronontheweb:
This fixes a bug I noticed while adding Akka.Hosting support to [Phobos](https://phobos.petabridge.com/)

https://twitter.com/Aaronontheweb/status/1512497162948595717 - this turned out to be caused by Akka.Hosting calling `ServiceCollection.BuildProvider()` in multiple instances - which causes the service registration delegates to be called multiple times, thus the version of a dependency your `ActorSystem` is running with can be different than the one ASP.NET might resolve.

## Changes

We've re-arranged the internals around Microsoft.Extensions.DependencyInjection inside Akka.Hosting to never actually instantiate the `IServiceProvider` ourselves - we let the runtime handle that so we end up with a consistent, single instance across the entire `IHost`.

We've also changed it such that the `AkkaHostedService` will terminate the entire application in the event of a failure at startup - whereas before it would simply error out in the background without saying anything.

## Checklist

### DIFF
--- a/src/Akka.Hosting/AkkaConfigurationBuilder.cs
+++ b/src/Akka.Hosting/AkkaConfigurationBuilder.cs
@@ -55,7 +55,8 @@ namespace Akka.Hosting
     /// </summary>
     internal sealed class SerializerRegistration
     {
-        public SerializerRegistration(string id, ImmutableHashSet<Type> typeBindings, Func<ExtendedActorSystem, Serializer> serializerFactory)
+        public SerializerRegistration(string id, ImmutableHashSet<Type> typeBindings,
+            Func<ExtendedActorSystem, Serializer> serializerFactory)
         {
             Id = id;
             TypeBindings = typeBindings;
@@ -63,9 +64,9 @@ namespace Akka.Hosting
         }
 
         public string Id { get; }
-        
+
         public Func<ExtendedActorSystem, Serializer> SerializerFactory { get; }
-        
+
         public ImmutableHashSet<Type> TypeBindings { get; }
     }
 
@@ -190,7 +191,8 @@ namespace Akka.Hosting
         }
 
         public AkkaConfigurationBuilder WithCustomSerializer(
-            string serializerIdentifier, IEnumerable<Type> boundTypes, Func<ExtendedActorSystem, Serializer> serializerFactory)
+            string serializerIdentifier, IEnumerable<Type> boundTypes,
+            Func<ExtendedActorSystem, Serializer> serializerFactory)
         {
             var serializerRegistration = new SerializerRegistration(serializerIdentifier,
                 boundTypes.ToImmutableHashSet(), serializerFactory);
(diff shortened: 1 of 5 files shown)


src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -190,47 +191,55 @@ public AkkaConfigurationBuilder StartActors(ActorStarter starter)
...
         {
             var serializerRegistration = new SerializerRegistration(serializerIdentifier,
                 boundTypes.ToImmutableHashSet(), serializerFactory);
-            _serializers.Add(serializerRegistration);
+            Serializers.Add(serializerRegistration);
             return this;
         }
 
-        internal void Build()
+        internal void Bind()
```
Aaronontheweb, line 203: Separate `Build` and `Bind` - need to have `Bind` run outside the context of the `IServiceProvider` construction in order to make sure that all of the registrations are added up front. We may not even really need the `Build` method any longer.

src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -190,47 +191,55 @@ public AkkaConfigurationBuilder StartActors(ActorStarter starter)
...
 
-                // start the IHostedService which will run Akka.NET
-                _serviceCollection.AddHostedService<AkkaHostedService>();
+            ServiceCollection.AddSingleton<ActorRegistry>(sp =>
+            {
+                return ActorRegistry.For(sp.GetRequiredService<ActorSystem>());
+            });
+        }
 
+        private  static Func<IServiceProvider, ActorSystem> ActorSystemFactory()
```
Aaronontheweb, line 214: The `ActorSystem` construction now uses the DI container to resolve the `AkkaConfigurationBuilder` in order to keep context boundaries clean. As a result most of the `private` members of the `AkkaConfigurationBuilder` are now `internal` so they can be accessed from this delegate.

src/Akka.Hosting/AkkaHostedService.cs
```diff
@@ -11,30 +14,46 @@ namespace Akka.Hosting
...
         private readonly AkkaConfigurationBuilder _configurationBuilder;
         private readonly IHostApplicationLifetime _hostApplicationLifetime;
+        private readonly ILogger<AkkaHostedService> _logger;
 
-        public AkkaHostedService(AkkaConfigurationBuilder configurationBuilder, IHostApplicationLifetime hostApplicationLifetime)
+        public AkkaHostedService(AkkaConfigurationBuilder configurationBuilder, IServiceProvider serviceProvider, ILogger<AkkaHostedService> logger)
         {
             _configurationBuilder = configurationBuilder;
-            _hostApplicationLifetime = hostApplicationLifetime;
+            _hostApplicationLifetime = serviceProvider.GetRequiredService<IHostApplicationLifetime>();
```
Aaronontheweb, line 25: I actually should change this back - came up during refactoring while we were working out instantiation issues.

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -50,14 +47,21 @@ public static IServiceCollection AddAkka(this IServiceCollection services, strin
         /// Starts a background <see cref="IHostedService"/> that runs the <see cref="ActorSystem"/>
         /// and manages its lifecycle in accordance with Akka.NET best practices.
         /// </remarks>
-        public static IServiceCollection AddAkka(this IServiceCollection services, string actorSystemName, Action<AkkaConfigurationBuilder, ServiceProvider> builder)
+        public static IServiceCollection AddAkka(this IServiceCollection services, string actorSystemName, Action<AkkaConfigurationBuilder, IServiceProvider> builder)
```
Aaronontheweb, line 50: Fixed the signature here - this is binary compatible.

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -50,14 +47,21 @@ public static IServiceCollection AddAkka(this IServiceCollection services, strin
...
         {
             var b = new AkkaConfigurationBuilder(services, actorSystemName);
-            var sp = services.BuildServiceProvider();
```
Aaronontheweb, line 56: This is the key fix - can't do this anymore.

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -50,14 +47,21 @@ public static IServiceCollection AddAkka(this IServiceCollection services, strin
...
-            builder(b, sp);
+            services.AddSingleton<AkkaConfigurationBuilder>(sp =>
+            {
+                
+                builder(b, sp);
```
Aaronontheweb, line 56: Run the builder inside DI resolution scope.

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -50,14 +47,21 @@ public static IServiceCollection AddAkka(this IServiceCollection services, strin
...
+                return b;
+            });
             
             // registers the hosted services and begins execution
-            b.Build();
+            b.Bind();
```
Aaronontheweb, line 61: Run the DI bindings exposed by Akka.NET outside the DI resolution scope so they're available as soon as the `IHost` starts.

Aaronontheweb (commented):
Reviewed my changes - found a few things I should clean up

src/Akka.Hosting/AkkaConfigurationBuilder.cs
Aaronontheweb: Removed the `Build` method as it's no longer needed.

PR #33, opened 2022-04-09
add `GetAsync` method to `ActorRegistry`

Aaronontheweb:
Somewhat of an experiment - considering adding this to allow non-Akka.NET consumers to `await` for an `IActorRef` to be added to the `ActorRegistry` in the future. I'll be adding tests for the `ActorRegistry` overall in a separate PR but I have tested this functionality and it does work. However, I don't want to get users into the habit of explicitly awaiting everything from the registry just because we have it - all actors started via a `WithActors` delegate are guaranteed to be created _and available in the `ActorRegistry`_ before your `IHost.StartAsync` method exits, so you shouldn't need to await the `ActorRegsitry` when fetching any top-level actors created by the builder itself.

_However_, if you needed a child actor of some kind - one that is started _after_ the `IHost.StartAsync` method completes then this might be useful.

## Changes

Enables entries to be added to the `ActorRegistry` later in the future (i.e. actors not started explicitly via Akka.Hosting's delegates - child actors typically) without forcing consumers of the registry to poll for changes.

## Checklist

### DIFF
--- a/src/Akka.Hosting/ActorRegistry.cs
+++ b/src/Akka.Hosting/ActorRegistry.cs
@@ -275,22 +275,9 @@ namespace Akka.Hosting
             {
                 // first step during timeout is to remove our registration
                 var d = (ConcurrentDictionary<Type, ImmutableHashSet<WaitForActorRegistration>>)dict;
-                var removed = false;
-                while (!removed)
-                {
-                    if(d.TryGetValue(key, out var registrations))
-                    {
-                        var original = registrations;
-                        registrations = registrations.Remove(waitingRegistration);
-                        if (d.TryUpdate(key, registrations, original))
-                            removed = true;
-                    }
-                    else // no key exists
-                    {
-                        removed = true;
-                    }
-                }
-                
+                d.AddOrUpdate(key, type => ImmutableHashSet<WaitForActorRegistration>.Empty,
+                    (type, set) => set.Remove(waitingRegistration));
+
                 // next, cancel the task
                 waitingRegistration.Waiter.TrySetCanceled(ct);
             };
--- a/src/Akka.Hosting/ActorRegistry.cs
+++ b/src/Akka.Hosting/ActorRegistry.cs
@@ -38,7 +38,7 @@ namespace Akka.Hosting
 
         /// <inheritdoc cref="IRequiredActor{TActor}.ActorRef"/>
         public IActorRef ActorRef { get; }
-    } 
+    }
 
     /// <summary>
     /// INTERNAL API
@@ -58,12 +58,10 @@ namespace Akka.Hosting
     {
         public ActorRegistryException(string message) : base(message)
         {
-            
         }
-        
+
         public ActorRegistryException(string message, Exception innerException) : base(message, innerException)
         {
-            
         }
     }
 
@@ -72,15 +70,12 @@ namespace Akka.Hosting
     /// </summary>
     public sealed class DuplicateActorRegistryException : ActorRegistryException
     {
-
         public DuplicateActorRegistryException(string message) : base(message)
         {
-            
         }
-        
+
         public DuplicateActorRegistryException(string message, Exception innerException) : base(message, innerException)
         {
-            
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -16,6 +19,41 @@ public override ActorRegistry CreateExtension(ExtendedActorSystem system)
         }
     }
 
+    /// <summary>
+    /// Used to implement "wait for actor" mechanics
+    /// </summary>
+    internal sealed class WaitForActorRegistration : IEquatable<WaitForActorRegistration>
```
Aaronontheweb, line 25: Registration handle - we create one of these per-waiter.

src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -29,6 +67,16 @@ public class ActorRegistry : IExtension
...
             new ConcurrentDictionary<Type, IActorRef>();
 
+        /// <summary>
+        /// In the event that an actor is not available yet, typically during the very beginning of ActorSystem startup,
+        /// we can wait on that actor becoming available.
+        /// </summary>
+        /// <remarks>
+        /// Have to store a collection of <see cref="WaitForActorRegistration"/>s here so each waiter gets its own cancellation token.
+        /// </remarks>
+        private readonly ConcurrentDictionary<Type, ImmutableHashSet<WaitForActorRegistration>> _actorWaiters =
```
Aaronontheweb, line 77: Contemplated just using a simple `lock` here but it became tricky once we started having the `CancellationTokenRegistration` delegates also competing to update entries. Opted for the compare-and-swap semantics of the `ConcurrentDictionary` instead.

> Man, this would have been so simple had I just encapsulated all of this inside an actor, but since we're trying to keep _this_ data structure from being dependent on the `ActorSystem` itself I went with this.

src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -84,6 +134,72 @@ public bool TryGet(Type key, out IActorRef actor)
...
+                }
+            }
+        }
+
+        private static Action<object> CancelWaiter(Type key, CancellationToken ct, WaitForActorRegistration waitingRegistration)
+        {
+            return dict =>
+            {
+                // first step during timeout is to remove our registration
+                var d = (ConcurrentDictionary<Type, ImmutableHashSet<WaitForActorRegistration>>)dict;
```
Aaronontheweb, line 181: We use a blend of closures and passing in state objects via parameters here - I think this is probably fine.

src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -53,6 +101,8 @@ public bool TryRegister(Type key, IActorRef actor, bool overwrite = false)
                 return _actorRegistrations.TryAdd(key, actor);
             else
                 _actorRegistrations[key] = actor;
+            
+            NotifyWaiters(key, _actorRegistrations[key]);
```
Aaronontheweb, line 105: Notify any waiters each time we push an update.

src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -84,6 +134,72 @@ public bool TryGet(Type key, out IActorRef actor)
...
+            {
+                return storedActor;
+            }
+
+            var tcs = new TaskCompletionSource<IActorRef>(TaskCreationOptions.RunContinuationsAsynchronously);
+            var waitingRegistration = new WaitForActorRegistration(key, tcs);
+            var registration = ct.Register(CancelWaiter(key, ct, waitingRegistration), _actorWaiters);
+            waitingRegistration.CancellationRegistration = registration;
+
+            _actorWaiters.AddOrUpdate(key,
```
Aaronontheweb, line 155: ConcurrentDictionary add semantics make this clean on additions and on removes... I think I'll clean this up for cancellations too.

Aaronontheweb (commented):
Added some design comments

Aaronontheweb:
Leaving this open for comments in the interim - doubles as a design discussion issue and feature implementation.

Aaronontheweb:
So I've finally seen a use-case where this feature is necessary out in the wild: user tried to access an actor in the `ActorRegistry` from a `BackgroundWorker` and got NRE'd. They added a `Task.Delay` to fix the issue - we should probably just include this async method to include a built-in way to wait for the actor to be created.

Arkatufus (approved):
LGTM

Aaronontheweb:
I'll address the merge conflict here

Aaronontheweb:
Rebased, but I need to add some additional specs for this feature first @Arkatufus

PR #50, opened 2022-05-19
added throw-able `ActorRegistry.Register` method

Aaronontheweb:
Wanted to do this to make assertions easier to implement within the `AkkaConfigurationBuilder`.

### DIFF
--- a/src/Akka.Hosting/ActorRegistry.cs
+++ b/src/Akka.Hosting/ActorRegistry.cs
@@ -44,8 +44,12 @@ namespace Akka.Hosting
 
         /// <inheritdoc cref="IActorRegistry.Register{TKey}"/>
         /// <exception cref="DuplicateActorRegistryException">Thrown when the same value is inserted twice and overwriting is not allowed.</exception>
+        /// <exception cref="ArgumentNullException">Thrown when a <c>null</c> <see cref="IActorRef"/> is registered.</exception>
         public void Register<TKey>(IActorRef actor, bool overwrite = false)
         {
+            if (actor == null)
+                throw new ArgumentNullException(nameof(actor), "Cannot register null actors");
+            
             if (!TryRegister<TKey>(actor, overwrite))
             {
                 throw new DuplicateActorRegistryException(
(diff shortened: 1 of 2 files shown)


src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -30,6 +42,17 @@ public class ActorRegistry : IActorRegistry, IExtension
...
             new ConcurrentDictionary<Type, IActorRef>();
 
+        /// <inheritdoc cref="IActorRegistry.Register{TKey}"/>
+        /// <exception cref="DuplicateActorRegistryException">Thrown when the same value is inserted twice and overwriting is not allowed.</exception>
+        public void Register<TKey>(IActorRef actor, bool overwrite = false)
+        {
+            if (!TryRegister<TKey>(actor, overwrite))
+            {
+                throw new DuplicateActorRegistryException(
+                    $"An actor for type {typeof(TKey)} has already been . Call `Register(IActorRef, bool overwrite=true)` to avoid this error or use a different key.");
```
Arkatufus, line 52: "has already been ." missing "registered"

src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -160,6 +183,13 @@ public interface IReadOnlyActorRegistry : IEnumerable<KeyValuePair<Type, IActorR
     /// </remarks>
     public interface IActorRegistry: IReadOnlyActorRegistry
     {
+        /// <summary>
+        /// Registers an actor into the registry. Throws an exception upon failure.
+        /// </summary>
+        /// <param name="actor">The bound <see cref="IActorRef"/>, if any. Is set to <see cref="ActorRefs.Nobody"/> if key is not found.</param>
```
Arkatufus, line 189: So we can register a null value to a key?

Arkatufus (changes requested):
Seems like there are some typo and ambiguity in the XML doc

src/Akka.Hosting/ActorRegistry.cs
Aaronontheweb: whoops - nice catch

src/Akka.Hosting/ActorRegistry.cs
Aaronontheweb: You can - nothing preventing that for now.

Aaronontheweb: Although this XML-DOC is wrong.

Arkatufus: Might want to do a null check and throw a `NullArgumentException`, maybe?

Aaronontheweb: Good idea

Arkatufus (approved):
LGTM

PR #66, opened 2022-06-15
Add Akka.Cluster.Tools.Client support

Arkatufus:
I'm following Microsoft lead on this one, using an option-like mutable POCO object as a configurator, then letting the user modify it through an action delegate.

~~Ready for review, but still need some unit tests to validate that all the settings are applied correctly.~~


Arkatufus:
Still need some specs

Arkatufus:
Do we need to make a sample project for this?

src/Akka.Cluster.Hosting/ClusterClientOptions.cs
```diff
@@ -0,0 +1,80 @@
...
+        /// How often failure detection heartbeat messages for detection of failed connections should be sent.
+        /// </summary>
+        public TimeSpan? HeartbeatInterval { get; set; }
+
+        /// <summary>
+        /// Number of potentially lost/delayed heartbeats that will be accepted before considering it to be an anomaly. 
+        /// The ClusterClient is using the <see cref="DeadlineFailureDetector"/>, which will trigger if there are 
+        /// no heartbeats within the duration <see cref="HeartbeatInterval"/> + <see cref="AcceptableHeartbeatPause"/>.
+        /// </summary>
+        public TimeSpan? AcceptableHeartbeatPause { get; set; }
```
Aaronontheweb, line 37: So one piece of feedback on this class, take it or leave it - I don't usually expose _every_ option on here. Only the 20% that 80% of people need (i.e. the `InitialContacts`.)

Might be a good idea to pare that down just to avoid people thinking they have to set all of these. Advanced users can always supply the HOCON instead. What do you think?

src/Akka.Cluster.Hosting/ClusterClientReceptionistOptions.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// will trigger if there are no heartbeats within the duration
+        /// heartbeat-interval + acceptable-heartbeat-pause, i.e. 15 seconds with
+        /// the default settings.
+        /// </summary>
+        public TimeSpan? AcceptableHeartbeatPause { get; set; }
+
+        /// <summary>
+        /// Failure detection checking interval for checking all ClusterClients
+        /// </summary>
+        public TimeSpan? FailureDetectionInterval { get; set; }
```
Aaronontheweb, line 49: Same feedback as above - I'd leave it as just Name, Role, NumberOfContracts, and ResponseTunnelReceiveTimeout.

On the last property I'd make it clear that this is optional and defaults to 60 seconds when not set.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -367,5 +372,47 @@ public static class AkkaClusterHostingExtensions
...
+
+        /// <summary>
+        /// Configures a <see cref="ClusterClientReceptionist"/> for the <see cref="ActorSystem"/>
+        /// </summary>
+        /// <param name="builder">The builder instance being configured.</param>
+        /// <param name="configure">An action that can be used to configure the receptionist settings</param>
+        /// <returns>The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.</returns>
+        public static AkkaConfigurationBuilder WithClusterClientReceptionist(
+            this AkkaConfigurationBuilder builder,
+            Action<ClusterClientReceptionistOptions> configure = null)
```
Aaronontheweb, line 384: Will this successfully start a `ClusterClientReceptionist` by default?

src/Akka.Cluster.Hosting/ClusterClientOptions.cs
```diff
@@ -0,0 +1,80 @@
...
+using Akka.Remote;
+
+namespace Akka.Cluster.Hosting
+{
+    public class ClusterClientOptions
+    {
+        /// <summary>
+        /// Actor paths of the <see cref="ClusterReceptionist"/> actors on the servers (cluster nodes) that the client will try to contact initially.
+        /// </summary>
+        public HashSet<ActorPath> InitialContacts { get; set; }
```
Aaronontheweb, line 15: Have this default to `/system/receptionist` - eliminates one big papercut that way.

src/Akka.Cluster.Hosting.Tests/ClusterClientSpecs.cs
```diff
@@ -0,0 +1,72 @@
...
+using FluentAssertions;
+using FluentAssertions.Extensions;
+using Xunit;
+
+namespace Akka.Cluster.Hosting.Tests;
+
+public class ClusterClientSpecs
+{
+    [Fact(DisplayName = "ClusterClientReceptionistSettings should be set correctly")]
+    public void ClusterClientReceptionistSettingsSpec()
```
Aaronontheweb, line 13: LGTM

src/Akka.Cluster.Hosting/ClusterClientOptions.cs
Aaronontheweb: Ah, I see - these have to be remote `ActorPath`s.... Any ideas how we can make that easier?

Aaronontheweb (commented):
I'm on the fence about the config object designs here - this is a very awkward to use area of Akka.NET that I would like to make easier to do right and I'm not sure we're there yet. What do you think @Arkatufus ?

src/Akka.Cluster.Hosting/ClusterClientOptions.cs
Arkatufus: Sounds good to me, we can ditch the options object altogether in that case.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Arkatufus: Yes, it would, everything would be at the default configuration if nothing is provided.

src/Akka.Cluster.Hosting/ClusterClientOptions.cs
Arkatufus: Not that I can see, the `ClusterClient` would need to know where the receptionist is to gather more contact informations.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -367,5 +373,86 @@ public static class AkkaClusterHostingExtensions
...
+        /// Configures a <see cref="ClusterClientReceptionist"/> for the <see cref="ActorSystem"/>
+        /// </summary>
+        /// <param name="builder">The builder instance being configured.</param>
+        /// <param name="name">Actor name of the ClusterReceptionist actor under the system path, by default it is /system/receptionist</param>
+        /// <param name="role">Checks that the receptionist only start on members tagged with this role. All members are used if empty.</param>
+        /// <returns>The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.</returns>
+        public static AkkaConfigurationBuilder WithClusterClientReceptionist(
+            this AkkaConfigurationBuilder builder,
+            string name = "receptionist",
+            string role = null)
```
Aaronontheweb, line 387: LGTM

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -367,5 +373,86 @@ public static class AkkaClusterHostingExtensions
...
+        /// to discover all of the receptionists in the cluster.
+        /// </para>
+        /// <para>
+        /// This should look something like "akka.tcp://systemName@networkAddress:2552/system/receptionist"
+        /// </para></param>
+        /// <typeparam name="TKey">The key type to use for the <see cref="ActorRegistry"/>.</typeparam>
+        /// <returns>The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.</returns>
+        public static AkkaConfigurationBuilder WithClusterClient<TKey>(
+            this AkkaConfigurationBuilder builder,
+            IList<ActorPath> initialContacts)
```
Aaronontheweb, line 422: Add an overload that takes `string`  instead of `ActorPath` - just to reduce some upfront friction there.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -367,5 +373,86 @@ public static class AkkaClusterHostingExtensions
...
+        /// <param name="initialContacts"> <para>
+        /// List of <see cref="ClusterClientReceptionist"/> <see cref="ActorPath"/> that will be used as a seed
+        /// to discover all of the receptionists in the cluster.
+        /// </para>
+        /// <para>
+        /// This should look something like "akka.tcp://systemName@networkAddress:2552/system/receptionist"
+        /// </para></param>
+        /// <typeparam name="TKey">The key type to use for the <see cref="ActorRegistry"/>.</typeparam>
+        /// <returns>The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.</returns>
+        public static AkkaConfigurationBuilder WithClusterClient<TKey>(
```
Aaronontheweb, line 420: I'm trying to imagine a scenario where there could be more than one `ClusterClient` but I guess it doesn't hurt to make the user specify one so they know where to look in the `ActorRegistry`.

Aaronontheweb (changes requested):
One minor change - should add a convenience overload that takes an `IList<string>` instead.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Aaronontheweb: Got it, unlikely but not impossible

Aaronontheweb (approved):
This looks excellent.

PR #72, opened 2022-07-01
Add ILoggerFactory logger support

### LINKED ISSUE #61: Add support for Akka.Logger.Serilog.Hosting
I'm successfully using the hosting package, but I still have to configure Serilog in hocon, and ensure that the Akka.Logger.Serilog package is added to the project.

Add a project for Akka.Logger.Serilog.Hosting that includes the dependency on Akka.Logger.Serilog, and has a builder extension method for WithSerilogLogger(string loglevel)

N/A

**Additional context**
A simple extension method could be added that sets the logger in the hocon

```csharp
public static class AkkaLoggerSerilogHostingExtensions
    {
        /// <summary>
        /// Configures Serilog logger.
        /// </summary>
        /// <param name="builder">The builder instance being configured.</param>
        /// <param name="loglevel">Specifies loglevel for <see cref="ActorSystem"/></param>
        /// <returns>The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.</returns>
        public static AkkaConfigurationBuilder WithSerilogLogger(this AkkaConfigurationBuilder builder, string loglevel)
        {
            if (!string.IsNullOrEmpty(loglevel))
            {
                builder.AddHocon($"akka {{ loglevel={loglevel},  loggers=[\"Akka.Logger.Serilog.SerilogLogger, Akka.Logger.Serilog\"]}}", HoconAddMode.Prepend);
            }

            return builder;
        }
    }
```

I have this code running, so let me know if you'd like me to send a PR.

Arkatufus:
Fixes #61

## Changes
- Add support to route Akka.NET logs to ILogger

### DIFF
--- a/src/Akka.Hosting/Logging/AkkaLoggerFactoryExtensions.cs
+++ b/src/Akka.Hosting/Logging/AkkaLoggerFactoryExtensions.cs
@@ -5,7 +5,10 @@
 // -----------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
+using System.Linq;
 using Akka.Actor;
+using Akka.Configuration;
 using Akka.Event;
 using Microsoft.Extensions.Logging;
 
@@ -13,19 +16,23 @@ namespace Akka.Hosting.Logging
 {
     public static class AkkaLoggerFactoryExtensions
     {
-        public static AkkaConfigurationBuilder WithLoggerFactory(
-            this AkkaConfigurationBuilder builder,
-            ILoggerFactory loggerFactory,
-            string timestampFormat = LoggerFactoryLogger.DefaultTimeStampFormat)
+        public static AkkaConfigurationBuilder WithLoggerFactory(this AkkaConfigurationBuilder builder)
         {
-            return builder.WithActors((system, registry) =>
-            {
-                var extSystem = (ExtendedActorSystem)system;
-                var logger = extSystem.SystemActorOf(
-                    props: Props.Create(() => new LoggerFactoryLogger(loggerFactory, timestampFormat)), 
-                    name: "log-ILoggerLogger");
-                logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TimeSpan.FromSeconds(5));
-            });
+            return builder.AddHocon("akka.loggers = [\"Akka.Hosting.Logging.LoggerFactoryLogger, Akka.Hosting\"]");
         }
+        
+        public static AkkaConfigurationBuilder AddLoggerFactory(this AkkaConfigurationBuilder builder)
+        {
+            var loggers = builder.Configuration.HasValue
+                ? builder.Configuration.Value.GetStringList("akka.loggers")
+                : new List<string>();
+            
+            if(loggers.Count == 0)
+                loggers.Add("Akka.Event.DefaultLogger");
+            
+            loggers.Add("Akka.Hosting.Logging.LoggerFactoryLogger, Akka.Hosting");
+            return builder.AddHocon($"akka.loggers = [{string.Join(", ", loggers.Select(s => $"\"{s}\""))}]");
+        }
+        
     }
 }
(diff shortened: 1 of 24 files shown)


src/Akka.Hosting/Logging/LoggerFactoryLogger.cs
```diff
@@ -0,0 +1,135 @@
...
+using Akka.Event;
+using Microsoft.Extensions.Logging;
+using LogLevel = Microsoft.Extensions.Logging.LogLevel;
+
+namespace Akka.Hosting.Logging
+{
+    public class LoggerFactoryLogger: ActorBase, IRequiresMessageQueue<ILoggerMessageQueueSemantics>
+    {
+        public const string DefaultTimeStampFormat = "yy/MM/dd-HH:mm:ss.ffff";
+        private const string DefaultMessageFormat = "[{{Timestamp:{0}}}][{{SourceContext}}][{{LogSource}}][{{ActorPath}}][{{Thread:0000}}]: {{Message}}";
```
Arkatufus, line 22: I'm not sure if this is the best way to format the string, I'm open to suggestions.

src/Akka.Hosting/Logging/AkkaLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+using System;
+using Akka.Actor;
+using Akka.Event;
+using Microsoft.Extensions.Logging;
+
+namespace Akka.Hosting.Logging
+{
+    public static class AkkaLoggerFactoryExtensions
+    {
+        public static AkkaConfigurationBuilder WithLoggerFactoryLogger(
```
Aaronontheweb, line 16: I'd rename this to `WithLoggerFactory`

Aaronontheweb (changes requested):
Overall looks really good, but:

- Needs some unit tests
- Have some other minor changes around class naming et al to consider.

src/Akka.Hosting/Logging/AkkaLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+            this AkkaConfigurationBuilder builder,
+            ILoggerFactory loggerFactory,
+            string timestampFormat = LoggerFactoryLogger.DefaultTimeStampFormat)
+        {
+            return builder.WithActors((system, registry) =>
+            {
+                var extSystem = (ExtendedActorSystem)system;
+                var logger = extSystem.SystemActorOf(
```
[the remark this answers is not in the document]
Aaronontheweb: Needs to be under `/system` or we lose all output from everything else once `/user` actors shutdown.

Aaronontheweb:
I'm also considering whether or not this should be instantiated through the normal `Logger` setup in Akka.NET in order to make sure this actor is kept alive when other `/system` actors are shutting down... That'd require:

1. A new `Setup` type which accepts the `ILoggerFactory`
2. A separate initiator actor that grabs that `Setup` and uses it to register the logger
3. A separate NuGet package.

Probably total overkill - as long as we unit test this setup **and add a sample that uses it inside Akka.Cluster** so Akka.Remote shutdown events can be observed, I think that should be fine. Does that make sense @Arkatufus ?

src/Akka.Hosting/Logging/AkkaLoggerFactoryExtensions.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace Akka.Hosting.Logging
+{
+    public static class AkkaLoggerFactoryExtensions
+    {
+        public static AkkaConfigurationBuilder WithLoggerFactory(this AkkaConfigurationBuilder builder)
+        {
+            return builder.AddHocon("akka.loggers = [\"Akka.Hosting.Logging.LoggerFactoryLogger, Akka.Hosting\"]");
+        }
+        
+        public static AkkaConfigurationBuilder AddLoggerFactory(this AkkaConfigurationBuilder builder)
```
Aaronontheweb, line 24: I want to rename this method but I honestly can't think of a better one. `AddMsftLogger` ?

Aaronontheweb (approved):
Looks good to me - we night need to change things down the road depending on user feedback, but I think we should ship this as-is.

PR #74, opened 2022-07-05
Nuke.Common

eaba:
## Changes

Nuke.Common from 6.0.1 to 6.1.2

### DIFF
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -26,8 +26,9 @@ using static Nuke.Common.Tools.Git.GitTasks;
 using Octokit;
 using Nuke.Common.Utilities;
 
-[CheckBuildProjectConfigurations]
 [ShutdownDotNetAfterServerBuild]
+[DotNetVerbosityMapping]
+[UnsetVisualStudioEnvironmentVariables]
 partial class Build : NukeBuild
 {
     /// Support plugins are available for:
--- a/build/Build.cs
+++ b/build/Build.cs
@@ -126,7 +126,6 @@ partial class Build : NukeBuild
           .Except(SourceDirectory.GlobFiles("**/*Tests.csproj", "**/*Tests*.csproj"));
           foreach (var project in projects)
           {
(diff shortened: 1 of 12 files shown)


build/Build.cs
```diff
@@ -158,6 +158,10 @@ partial class Build : NukeBuild
                 DotNetNuGetPush(s => s
                  .SetTimeout(TimeSpan.FromMinutes(10).Minutes)
                  .SetTargetPath(package)
+                 //.SetRuntime(platform)
```
Aaronontheweb, line 161: If we don't need these, please remove them rather than leave them uncommented.

build/Build.cs
```diff
@@ -175,28 +183,31 @@ partial class Build : NukeBuild
     });
     Target AuthenticatedGitHubClient => _ => _
         .Unlisted()
-        .OnlyWhenDynamic(() => !string.IsNullOrWhiteSpace(GitHubToken))
+        .OnlyWhenDynamic(() => !string.IsNullOrWhiteSpace(GitHubActions.Token))
```
Aaronontheweb, line 186: LGTM

src/Directory.Build.props
```diff
@@ -220,8 +220,9 @@
     <TestsNetCoreFramework>net6.0</TestsNetCoreFramework>
     <XunitVersion>2.4.1</XunitVersion>
     <TestSdkVersion>17.2.0</TestSdkVersion>
+	<XunitRunneVisualstudio>2.4.5</XunitRunneVisualstudio>
     <AkkaVersion>1.4.39</AkkaVersion>
-    <MicrosoftExtensionsVersion>[3.0.0,)</MicrosoftExtensionsVersion>
```
Aaronontheweb, line 224: Need to leave the original behavior here, as this has upstream breaking changes for some applications still running older versions of .NET Core / .NET 5.

Aaronontheweb (changes requested):
Need a small amount of cleanup here.

PR #81, opened 2022-07-19
Fix logger factory logger

### LINKED ISSUE #78: WithLoggerFactory unable to load logger
**Version Information**
Akka 1.4.39
Akka.hosting 0.4.0

**Describe the bug**
Using `WithLoggerFactory` throws Akka.Configuration.ConfigurationException:

**To Reproduce**
Create a new console application using the dotnet cli
```
> dotnet new console
```

Install dependencies
```xml
    <ItemGroup>
      <PackageReference Include="Akka" Version="1.4.39" />
      <PackageReference Include="Akka.Hosting" Version="0.4.0" />
      <PackageReference Include="Microsoft.Extensions.Hosting" Version="3.0.0" />
    </ItemGroup>
```
Update program.cs
```csharp
using var host = CreateHostBuilder(args)
    .UseConsoleLifetime().Build();

await host.StartAsync();

var actorSystem = host.Services.GetRequiredService<ActorSystem>();

Console.ReadLine();

await actorSystem.Terminate();
await actorSystem.WhenTerminated;
await host.StopAsync();

static IHostBuilder CreateHostBuilder(string[] args) =>
    Host.CreateDefaultBuilder(args)
        .ConfigureServices(services =>
        {
            services.AddAkka("AkkaSandbox", (configBuilder, provider) =>
            {
                configBuilder.AddHocon("akka.loglevel = DEBUG")
                    .WithLoggerFactory();
            });
        });
```

Error message
```
crit: Akka.Hosting.AkkaHostedService[0]
      Unable to start AkkaHostedService - shutting down application
Akka.Configuration.ConfigurationException: Logger [Akka.Hosting.Logging.LoggerFactoryLogger, Akka.Hosting] specified in config cannot be loaded: System.AggregateException: One or more errors occurred. (Timeout after 00:00:05 seconds)
 ---> Akka.Actor.AskTimeoutException: Timeout after 00:00:05 seconds
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at Akka.Event.LoggingBus.AddLogger(ActorSystemImpl system, Type loggerType, LogLevel logLevel, String loggingBusName, TimeSpan timeout)
   at Akka.Event.LoggingBus.StartDefaultLoggers(ActorSystemImpl system)
 ---> System.AggregateException: One or more errors occurred. (Timeout after 00:00:05 seconds)
 ---> Akka.Actor.AskTimeoutException: Timeout after 00:00:05 seconds
   --- End of inner exception stack trace ---
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task`1.GetResultCore(Boolean waitCompletionNotification)
   at System.Threading.Tasks.Task`1.get_Result()
   at Akka.Event.LoggingBus.AddLogger(ActorSystemImpl system, Type loggerType, LogLevel logLevel, String loggingBusName, TimeSpan timeout)
   at Akka.Event.LoggingBus.StartDefaultLoggers(ActorSystemImpl system)
   --- End of inner exception stack trace ---
   at Akka.Event.LoggingBus.StartDefaultLoggers(ActorSystemImpl system)
   at Akka.Actor.LocalActorRefProvider.Init(ActorSystemImpl system)
   at Akka.Actor.Internal.ActorSystemImpl.Start()
   at Akka.Actor.ActorSystem.CreateAndStartSystem(String name, Config withFallback, ActorSystemSetup setup)
   at Akka.Actor.ActorSystem.Create(String name, ActorSystemSetup setup)
   at Akka.Hosting.AkkaConfigurationBuilder.<>c.<ActorSystemFactory>b__28_0(IServiceProvider sp)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitFactory(FactoryCallSite factoryCallSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSiteMain(ServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitCache(ServiceCallSite callSite, RuntimeResolverContext context, ServiceProviderEngineScope serviceProviderEngine, RuntimeResolverLock lockType)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite singletonCallSite, RuntimeResolverContext context)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteVisitor`2.VisitCallSite(ServiceCallSite callSite, TArgument argument)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.Resolve(ServiceCallSite callSite, ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.DynamicServiceProviderEngine.<>c__DisplayClass1_0.<RealizeService>b__0(ServiceProviderEngineScope scope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngine.GetService(Type serviceType, ServiceProviderEngineScope serviceProviderEngineScope)
   at Microsoft.Extensions.DependencyInjection.ServiceLookup.ServiceProviderEngineScope.GetService(Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
   at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
   at Akka.Hosting.AkkaHostedService.StartAsync(CancellationToken cancellationToken)

```

**Environment**
MacOS Monetery, .NET 6

Arkatufus:
Fixes #78

## Changes

This changes the API used to configure loggers for Akka.Hosting, we will need to deprecate the previous nuget release.
With the new API, you use the `ConfigureLoggers(Action<LoggerSetup>)` extension method to configure the Akka.NET logging system. The new API is a lot more flexible compared to the previous one and should be compatible with our other logger plug-ins.

```
builder.Services.AddAkka("MyActorSystem", configurationBuilder =>
{
    configurationBuilder
        .ConfigureLoggers(setup =>
        {
            // Example: This sets the minimum log level
            setup.LogLevel = LogLevel.DebugLevel;
            
            // Example: Clear all loggers
            setup.ClearLoggers();
            
            // Example: Add the default logger
            // NOTE: You can also use setup.AddLogger<DefaultLogger>();
            setup.AddDefaultLogger();
            
            // Example: Add the ILoggerFactory logger
            // NOTE:
            //   - You can also use setup.AddLogger<LoggerFactoryLogger>();
            //   - To use a specific ILoggerFactory instance, you can use setup.AddLoggerFactory(myILoggerFactory);
            setup.AddLoggerFactory();
        })
        .WithActors((system, registry) =>
        {
            var echo = system.ActorOf(act =>
            {
                act.ReceiveAny((o, context) =>
                {
                    Logging.GetLogger(context.System, "echo").Info($"Actor received {o}");
                    context.Sender.Tell($"{context.Self} rcv {o}");
                });
            }, "echo");
            registry.TryRegister<Echo>(echo); // register for DI
        });
});
```


src/Akka.Hosting/Logging/LoggerFactorySetup.cs
```diff
@@ -0,0 +1,22 @@
...
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using Akka.Actor.Setup;
+using Microsoft.Extensions.Logging;
+
+namespace Akka.Hosting.Logging
+{
+    public class LoggerFactorySetup : Setup
```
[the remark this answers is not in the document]
Aaronontheweb: That's fine.

src/Akka.Hosting/Logging/LoggerSetup.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        /// <summary>
+        /// Register a logger
+        /// </summary>
+        /// <returns>This <see cref="LoggerSetup"/> instance</returns>
+        public LoggerSetup AddLogger<T>() where T: ActorBase
+        {
+            var logger = typeof(T);
+            var methodInfo = logger.GetMethod(
```
Arkatufus, line 62: This reflection is used to get the static "magic method" called `ConfigureHosting` inside the logger class, if we found this static method, we will call it so the logger can register itself properly with Akka.Hosting

src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -94,6 +94,13 @@ public sealed class AkkaConfigurationBuilder
         /// Used to hold a reference to the <see cref="Sys"/> being started.
         /// </summary>
         internal Option<ActorSystem> Sys { get; set; } = Option<ActorSystem>.None;
+        
+        /// <summary>
+        /// INTERNAL API.
+        ///
+        /// Used to hold a reference to the dependency injection <see cref="ServiceProvider"/> being used to setup the <see cref="ActorSystem"/>.
+        /// </summary>
+        internal Option<IServiceProvider> ServiceProvider { get; set; } = Option<IServiceProvider>.None;
```
Aaronontheweb, line 103: We don't need to hold onto this here - it will be passed back into us in ` ActorSystemFactory()`

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -53,7 +54,7 @@ public static IServiceCollection AddAkka(this IServiceCollection services, strin
             var b = new AkkaConfigurationBuilder(services, actorSystemName);
             services.AddSingleton<AkkaConfigurationBuilder>(sp =>
             {
-                
+                b.ServiceProvider = new Option<IServiceProvider>(sp);
```
Aaronontheweb, line 57: No need for this - same comment as before.

src/Akka.Hosting/Logging/LoggerFactoryLogger.cs
```diff
@@ -10,28 +10,51 @@
...
         private readonly ILoggingAdapter _log = Akka.Event.Logging.GetLogger(Context.System.EventStream, nameof(LoggerFactoryLogger));
-        private ILoggerFactory _loggerFactory;
+        private readonly ILoggerFactory _loggerFactory;
         private readonly string _messageFormat;
 
         public LoggerFactoryLogger()
         {
             _messageFormat = string.Format(DefaultMessageFormat, DefaultTimeStampFormat);
+            var setup = Context.System.Settings.Setup.Get<LoggerFactorySetup>();
+            if (!setup.HasValue)
```
Aaronontheweb, line 53: LGTM

src/Akka.Hosting/Logging/LoggerSetup.cs
```diff
@@ -0,0 +1,108 @@
...
+using System.Linq;
+using System.Reflection;
+using System.Text;
+using Akka.Actor;
+using Akka.Configuration;
+using Akka.Event;
+
+namespace Akka.Hosting.Logging
+{
+    public class LoggerSetup
```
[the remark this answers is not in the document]
Aaronontheweb: Mark it as `sealed`

Aaronontheweb (changes requested):
Looks good so far, but add some unit tests.

src/Akka.Hosting/Logging/LoggingExtensions.cs
```diff
@@ -0,0 +1,69 @@
...
+    {
+        /// <summary>
+        /// Fluent interface to configure the Akka.NET logger system
+        /// </summary>
+        /// <param name="builder">The <see cref="AkkaConfigurationBuilder"/> being configured</param>
+        /// <param name="configurator">An action that can be used to modify the logging configuration</param>
+        /// <returns>The original <see cref="AkkaConfigurationBuilder"/> instance</returns>
+        public static AkkaConfigurationBuilder ConfigureLoggers(this AkkaConfigurationBuilder builder, Action<LoggerSetup> configurator)
+        {
+            var setup = new LoggerSetup(builder);
```
Aaronontheweb, line 24: Is there a way we can avoid taking a dependency on the `AkkaConfigurationBuilder` inside of here?

src/Akka.Hosting/AkkaConfigurationBuilder.cs
Arkatufus: Ww actually do, this will be used by extension methods to pull required services from the builder class

Aaronontheweb: This is pulled when we create the `LoggerSetup` right?

Aaronontheweb: Because if that's the case, no we don't - look at how we create the `DependencyResolverSetup` for instance:

https://github.com/akkadotnet/Akka.Hosting/blob/2f1683dd42e237badd823bb2ff4739d5efe51020/src/Akka.Hosting/AkkaConfigurationBuilder.cs#L227-L239

Arkatufus: The ILoggerFactory needs the service provider so that it can do an ILoggerFactory lookup _before_ the actor system starts, extension methods only have access to the builder instance, this is the only generic way I can think of for doing that, instead of trying to do a hook inside the actor system factory itself

Arkatufus: another way of doing this is to ask the user to provide the service provider instance for the ILoggerFactory logger extension method, which is the thing i'm trying hard to avoid

Arkatufus: The way the ILoggerFactory logger cheats the DI system is to do a lookup before the actor system starts, put it inside a Setup class, and pulls that Setup class in its constructor.
The big problem is that there is no way to access the ServiceProvider instance from inside an extension method, hence we have to provide that inside the builder instance itself

src/Akka.Hosting/Logging/LoggerSetup.cs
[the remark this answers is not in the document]
Arkatufus: Done

src/Akka.Hosting/AkkaConfigurationBuilder.cs
Aaronontheweb: I fixed this in a subsequent commit. Didn't need to change any of the public APIs.

src/Akka.Hosting/Logging/LoggingExtensions.cs
Aaronontheweb: I fixed this.

src/Akka.Hosting/Logging/LoggerSetup.cs
```diff
@@ -0,0 +1,108 @@
...
+using Akka.Event;
+
+namespace Akka.Hosting.Logging
+{
+    public sealed class LoggerSetup
+    {
+        private readonly List<Type> _loggers = new List<Type> { typeof(DefaultLogger) };
+        internal AkkaConfigurationBuilder Builder { get; }
+
+        internal LoggerSetup(AkkaConfigurationBuilder builder)
```
Aaronontheweb, line 23: I thought this was, originally a `Setup` class from Akka.NET - now that I know it's not, fine having it take a dependency on `AkkaConfigurationBuilder`.

src/Akka.Hosting/Logging/LoggerSetup.cs
Aaronontheweb: I think I still need to fix this - it's not covered by unit tests at the moment.

Aaronontheweb: Actually, looks like we don't need it.

Aaronontheweb (approved):
Removed all of the objectionable stuff and the tests still pass. Do we need to add more @Arkatufus ?

Arkatufus:
The reflection magic is needed for the `LoggerFactoryLogger` to support the `.AddLogger<LoggerFactoryLogger>()` feature, are we sure we want to remove that?

Aaronontheweb:
> The reflection magic is needed for the `LoggerFactoryLogger` to support the `.AddLogger<LoggerFactoryLogger>()` syntax, are we sure we want to remove that?

Added test coverage to see if this was a problem and then added a check into the `AkkaConfigurationBuilder` to automatically include any dependencies it needs.

src/Akka.Hosting.Tests/Logging/TestLogger.cs
```diff
@@ -0,0 +1,96 @@
+﻿using System;
+using System.Collections.Generic;
+using Microsoft.Extensions.Logging;
+using Xunit.Abstractions;
+
+namespace Akka.Hosting.Tests.Logging;
+
+public class TestLogger : ILogger
```
Aaronontheweb, line 8: Moved this into a shared class

src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -231,6 +233,22 @@ internal void Bind()
                 /*
                  * Build setups
                  */
+                
+                // check to see if we need a LoggerSetup
+                var hasLoggerSetup = config.Setups.Any(c => c is LoggerFactorySetup);
+                if (!hasLoggerSetup)
+                {
+                    var logger = sp.GetService<ILoggerFactory>();
```
Aaronontheweb, line 241: If the `ILoggerFactory` is available in the `IServiceProvider`, add the setup anyway. Costs nothing.

src/Akka.Hosting/LoggerConfigBuilder.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using Akka.Configuration;
+using Akka.Dispatch;
+using Akka.Event;
+
+namespace Akka.Hosting
+{
+    public sealed class LoggerConfigBuilder
```
Aaronontheweb, line 17: Moved into the `Akka.Hosting` namespace so it's automatically discoverable

src/Akka.Hosting/Logging/LoggerFactoryLogger.cs
```diff
@@ -26,12 +26,18 @@ public class LoggerFactoryLogger: ActorBase, IRequiresMessageQueue<ILoggerMessag
 
         private readonly ConcurrentDictionary<Type, ILogger> _loggerCache = new ConcurrentDictionary<Type, ILogger>();
```
Aaronontheweb, line 27: This does not need to be concurrent, still.

src/Akka.Hosting/LoggingExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using Akka.Event;
+using Akka.Hosting.Logging;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+namespace Akka.Hosting
```
Aaronontheweb, line 13: Moved into Akka.Hosting namespace for discoverability.

Aaronontheweb (commented):
Detailed some of my own changes

src/Akka.Hosting/AkkaConfigurationBuilder.cs
Aaronontheweb: This eliminates the need for any reflection magic and this scenario is covered by the `End2EndSpecs` I added.

Arkatufus: Oh, makes sense, good idea

src/Akka.Hosting/Logging/LoggerFactoryLogger.cs
Arkatufus: Done

Aaronontheweb:
Simplified the internals of the `LoggingFactoryLogger` so it only creates a single `ILogger<ActorSystem>` instance. The output logs look like this:

```
[Information] [22/07/21-15:15:26.9254][ActorSystem(MySys)][akka://test/system/testActor1][0023]: foo
```

Arkatufus:
LGTM, merging

PR #90, opened 2022-08-10
Add StartupTask feature

Arkatufus:
## Changes

Adds `AddStartup` using `StartupTask` delegate to the configuration builder. 
This feature is useful when a user need to run a specific initialization code if anf only if the `ActorSystem` and all of the actors have been started. Although it is semantically the same as `AddActors` and `WithActors`, it disambiguate the use-case with a guarantee that it will only be executed after everything is ready.

For example, this feature is useful for: 
- kicking off actor initializations by using `Tell()`s once all of the actor infrastructure are in place, or
- pre-populating certain persistence data after everything is set up and running, useful for unit testing.


Aaronontheweb (approved):
LGTM

PR #91, opened 2022-08-10
Add `WithExtensions` support

Arkatufus:
## Changes
Add `WithExtensions` support that simplifies adding `akka.extensions` HOCON setting


Aaronontheweb (approved):
LGTM

Aaronontheweb:
What happens if this method gets called multiple times?

PR #92, opened 2022-08-11
Improve WithExtensions so it can be used universally

Arkatufus:
## Changes
Improve `WithExtensions` so that it can:
- Be called multiple times without overriding each other
- Be used together along side manually declared extensions in HOCON


Aaronontheweb (approved):
LGTM

PR #95, opened 2022-08-24
Add split brain resolver configuration support to Akka.Cluster.Hosting

Arkatufus:
Adds additional optional parameter to `WithClustering()` to also set up a split brain resolver for the cluster.

### DIFF
--- a/src/Akka.Cluster.Hosting.Tests/SplitBrainResolverSpecs.cs
+++ b/src/Akka.Cluster.Hosting.Tests/SplitBrainResolverSpecs.cs
@@ -1,0 +1,178 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="SplitBrainResolverSpecs.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Cluster.Hosting.SBR;
+using Akka.Cluster.Hosting.Tests.Lease;
+using Akka.Cluster.SBR;
+using Akka.Hosting;
+using Akka.Remote.Hosting;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using Xunit;
+using Xunit.Abstractions;
+using LogLevel = Microsoft.Extensions.Logging.LogLevel;
+
+namespace Akka.Cluster.Hosting.Tests;
+
+public class SplitBrainResolverSpecs
+{
+    private readonly ITestOutputHelper _output;
+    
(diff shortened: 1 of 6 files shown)


Arkatufus:
Still need unit tests

Aaronontheweb:
> Adds additional optional parameter to `WithClustering()` to also set up a split brain resolver for the cluster.

this is excellent - one of the last things I need HOCON for currently. Let's get you some API feedback before you go through and add tests.

src/Akka.Cluster.Hosting/SBR/SplitBrainResolverOption.cs
```diff
@@ -0,0 +1,99 @@
...
+    /// Best effort is to keep the side that has most nodes, i.e. the majority side.
+    /// This is achieved by adding a delay before trying to acquire the lease on the
+    /// minority side.
+    /// </summary>
+    public sealed class LeaseMajorityOption : SplitBrainResolverOption
+    {
+        /// <summary>
+        /// A class type that extends the abstract class <see cref="Lease"/>
+        /// </summary>
+        public Type LeaseImplementation { get; set; }
```
Aaronontheweb, line 89: Do you want to refer to the concrete type itself here and extract the FQN later? That way we can statically type this.

Aaronontheweb (changes requested):
Overall idea looks good but I think you need to add an example and a unit test to see if the API "feels" good.

src/Akka.Cluster.Hosting/SBR/SplitBrainResolverOption.cs
Arkatufus: I don't think we can do that since this type is being instantiated inside `Akka.Coordination` and there's no good way for user to instantiate it during setup. I could add a helper function `WithLeaseImplementation<T>() where T: Lease` to help typing.

Aaronontheweb: Yeah a generic method like that may not be a bad idea - that's why I think it's worth playing with this SBR API some more to see how it feels.

Arkatufus: Scrapped this, I was wrong, `LeaseMajority.LeaseImplementation` is an absolute HOCON path to the lease configuration block, not the implementation class type.

src/Akka.Cluster.Hosting.Tests/SplitBrainResolverSpecs.cs
```diff
@@ -0,0 +1,178 @@
...
+        {
+            builder.WithClustering(sbrOptions: SplitBrainResolverOption.Default);
+        });
+
+        var system = host.Services.GetRequiredService<ActorSystem>();
+        
+        Cluster.Get(system).DowningProvider.Should().BeOfType<SplitBrainResolverProvider>();
+        var settings = new SplitBrainResolverSettings(system.Settings.Config);
+        settings.DowningStrategy.Should().Be(SplitBrainResolverSettings.KeepMajorityName);
+        settings.KeepMajorityRole.Should().BeNull();
```
Aaronontheweb, line 90: LGTM

Aaronontheweb (approved):
LGTM - we talked about the `Lease` API and decided that this is good enough for the time being.

PR #97, opened 2022-08-29
Add static typed `WithExtension` method

Arkatufus:
## Changes
- Add `WithExtension<T>()` method with type constraint and checks and that the type is not abstract or static
- `WithExtensions` now checks that the type parameter is derived from `IExtensionId` and that the type is not abstract or static


Aaronontheweb (approved):
LGTM

PR #99, opened 2022-08-30
Add Akka.Cluster.Hosting documentation

Arkatufus:


src/Akka.Cluster.Hosting/Akka.Cluster.Hosting.csproj
```diff
@@ -6,10 +6,6 @@
     <LangVersion>9</LangVersion>
   </PropertyGroup>
 
-  <ItemGroup>
-    <None Include="..\..\README.md" Pack="true" PackagePath="\" />
```
Aaronontheweb, line 10: Got it - this will use the new project-specific `README.md` instead

Aaronontheweb (approved):
LGTM

PR #102, opened 2022-09-06
Add Akka.Hosting.TestKit module

Arkatufus:
A proof of concept implementation of Akka.Hosting test kit.
Comments and reviews are welcome.

## Configuring the spec

The biggest difference is that, since the test is started asynchronously, the TestKit properties and methods would not be available in the unit test class constructor anymore. Since the spec depends on Microsoft HostBuilder, configuration has to be broken down into steps. There are overridable methods that user can use to override virtually all of the setup process. 

The steps of what overridable methods gets called is:
1. `ConfigureLogging(ILoggingBuilder)` -- Add custom logger and filtering rules here.
2. `ConfigureHostConfiguration(IConfigurationBuilder)` -- Inject any additional hosting environment configuration here, such as faking environment variables.
3. `ConfigureAppConfiguration(HostBuilderContext, IConfigurationBuilder)` -- Inject the application configuration here.
4. `ConfigureServices(HostBuilderContext, IServiceCollection)` -- Add additional services needed by the test here, such as mocked up services used inside the dependency injection.
5. User defined HOCON configuration is injected by overriding the `Config` property, it is not passed as part of the constructor anymore.
6. `ConfigureAkka(AkkaConfigurationBuilder, IServiceProvider)` -- This is called inside `AddAkka`, use this to configure the `ActorSystem`
7. There is a new `BeforeTestStart()` method. This method is called after the TestKit is initialized. Move all of the codes that used to belong in the constructor here.

`Akka.Hosting.TestKit` extends `Akka.TestKit.Xunit2` directly, all testing methods are available out of the box.
All of the properties, such as `Sys` and `TestActor` will be available once the unit test class is invoked.

## NOTES
~~Would not compile until https://github.com/akkadotnet/akka.net/pull/6093 is merged and released~~


Arkatufus:
Would not compile until https://github.com/akkadotnet/akka.net/pull/6093 is merged and released

Arkatufus:
Tests are passing

src/Akka.Hosting.TestKit.Tests/HostingSpecSpec.cs
```diff
@@ -0,0 +1,64 @@
...
+    {
+        private enum Echo
+        { }
+
+        public HostingSpecSpec(ITestOutputHelper output)
+            : base(nameof(HostingSpecSpec), output, logLevel: LogLevel.Debug)
+        {
+        }
+
+        protected override Task ConfigureAkka(AkkaConfigurationBuilder builder, IServiceProvider provider)
```
Aaronontheweb, line 29: How would I populate the `IServiceProvider` in advance?

src/Akka.Hosting.TestKit/TestKit.cs
```diff
@@ -0,0 +1,214 @@
...
+                });
+            });
+        }
+
+        protected virtual Config? Config { get; } = null;
+        
+        protected virtual void ConfigureLogging(ILoggingBuilder builder)
+        { }
+
+        protected virtual Task ConfigureAkka(AkkaConfigurationBuilder builder, IServiceProvider provider)
```
Aaronontheweb, line 119: Let's make this `abstract` - force the user to spell out what they want to use as part of their tests.

Aaronontheweb (changes requested):
By way of example, let's refactor at least 1 of the existing test suites in Akka.Hosting to use the `TestKit`. If there are issues with friction / the ease of use of the API that would be a valuable way to discover it before we push this out to end users.

Aaronontheweb:
I think https://github.com/akkadotnet/Akka.Hosting/tree/dev/src/Akka.Persistence.Hosting.Tests is a good candidate for testing out the new Akka.Hosting TestKit - not too big, reasonably complex use cases.

src/Akka.Hosting.TestKit.Tests/HostingSpecSpec.cs
Arkatufus: Override the `ConfigureServices` method

Arkatufus:
One snag that I've seen so far: This TestKit can only be used if a spec uses the same hosting configuration for all its unit test. If the unit tests were testing different hosting configurations, then the only way to use the test kit is to split each unit tests into classes.

Aaronontheweb (approved):

PR #107, opened 2022-09-12
Extend Persistence.SqlServer support with full fledged option configuration

### LINKED ISSUE #105: Add support for modifying the sql settings in the `WithSqlServerPersistence` method
We are unable to modify the schema-name, table-name and auto-initialize values when I use the `.WithSqlServerPersistence()` extension method.

We would like to be able to modify the schema-name, table-name and auto-initialize values for both the Journal and Snapshot tables when I use the `.WithSqlServerPersistence()` extension method.

We've tried adding a hocon config for this, like so:
```
.AddHocon(@"akka.persistence {
	                journal {
                        plugin = ""akka.persistence.journal.sql-server""
	                    sql-server {
                            schema-name = akka
                            auto-initialize = on
                        }
                    }
	                snapshot-store {
                        plugin = ""akka.persistence.snapshot-store.sql-server""
		                sql-server {
                            schema-name = akka
                            auto-initialize = on
                        }
                    }
                }");
```

But that does not work.  The tables are still initialized with `dbo` as the schema name.

Arkatufus:
Fixes #105

## Changes
Add support to use option classes to configure Akka.Persistence.SqlServer.Hosting

### DIFF
--- a/src/Akka.Hosting/HoconExtensions.cs
+++ b/src/Akka.Hosting/HoconExtensions.cs
@@ -1,0 +1,49 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="StringExtensions.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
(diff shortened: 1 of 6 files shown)


Aaronontheweb (changes requested):
Need to add some unit tests here to ensure that these changes are effective

Aaronontheweb (approved):
LGTM

PR #108, opened 2022-09-15
WithRemoting should only override declared values

Arkatufus:
Make `WithRemoting` write the minimum HOCON override possible to prevent unexpected overrides of previously declared HOCON overrides


Aaronontheweb (approved):
LGTM

PR #110, opened 2022-09-19
Spec Proposal to standardize HOCON path based configuration pattern

Arkatufus:
This will be used to standardize how Akka.Hosting addresses a very common HOCON configuration pattern.
This allows for a HOCON-less programmatical setup replacement for the HOCON path used to configure the HOCON property. 

The pattern:
```text
# This HOCON property references to a config block below
akka.discovery.method = akka.discovery.config

akka.discovery.config {
    class = "Akka.Discovery.Config.ConfigServiceDiscovery, Akka.Discovery"
    # other options goes here
}
```

Example implementation:
```csharp
// The base class for the option
public abstract class DiscoveryOptionBase : IOption
{ }

// The actual option implementation
public class ConfigOption : DiscoveryOptionBase
{
    // Actual option implementation here
    public Type Class => typeof(ConfigServiceDiscovery);

    public string ConfigPath => "akka.discovery.config";

    public void Apply(AkkaConfigurationBuilder builder, Setup setup = null)
    {
        // Modifies Akka.NET configuration either via HOCON or setup class
        builder.AddHocon($"akka.discovery.method = {ConfigPath}", HoconAddMode.Prepend);

        // Rest of configuration goes here
    }
}

 // Akka.Hosting extension implementation
 public static AkkaConfigurationBuilder WithDiscovery(
     this AkkaConfigurationBuilder builder,
     DiscoveryOptionBase discOption)
 {
     var setup = new DiscoverySetup();

     // gets called here
     discOption.Apply(builder, setup);
 }        
```

### DIFF
--- a/src/Akka.Cluster.Hosting/SBR/SplitBrainResolverOption.cs
+++ b/src/Akka.Cluster.Hosting/SBR/SplitBrainResolverOption.cs
@@ -5,12 +5,14 @@
 // -----------------------------------------------------------------------
 
 using System;
-using Akka.Annotations;
-using Akka.Coordination;
+using System.Text;
+using Akka.Actor.Setup;
+using Akka.Cluster.SBR;
+using Akka.Hosting;
 
 namespace Akka.Cluster.Hosting.SBR
 {
-    public abstract class SplitBrainResolverOption
+    public abstract class SplitBrainResolverOption: IOption
     {
         public static readonly SplitBrainResolverOption Default = new KeepMajorityOption();
         
@@ -18,6 +20,12 @@ namespace Akka.Cluster.Hosting.SBR
         /// if the <see cref="Role"/> is defined the decision is based only on members with that <see cref="Role"/>
         /// </summary>
         public string Role { get; set; }
+
+        public abstract string ConfigPath { get; }
+
+        public Type Class => typeof(SplitBrainResolverProvider);
+
+        public abstract void Apply(AkkaConfigurationBuilder builder, Setup setup = null);
     }
 
     /// <summary>
@@ -38,10 +46,37 @@ namespace Akka.Cluster.Hosting.SBR
     /// </summary>
     public sealed class StaticQuorumOption : SplitBrainResolverOption
     {
+        public override string ConfigPath => SplitBrainResolverSettings.StaticQuorumName;
+        
         /// <summary>
         /// Minimum number of nodes that the cluster must have
         /// </summary>
-        public int QuorumSize { get; set; } = 0;
+        public int? QuorumSize { get; set; } = 0;
+        
+        public override void Apply(AkkaConfigurationBuilder builder, Setup setup = null)
+        {
+            var sb = new StringBuilder("akka.cluster {");
+            sb.AppendLine($"downing-provider-class = \"{Class.AssemblyQualifiedName}\"");
+            sb.AppendLine("split-brain-resolver {");
+            sb.AppendLine($"active-strategy = {ConfigPath}");
+
+            var innerSb = new StringBuilder();
+            if (Role != null)
+                innerSb.AppendLine($"role = {Role}");
+            if(QuorumSize != null)
+                innerSb.AppendLine($"quorum-size = {QuorumSize}");
+
+            if (innerSb.Length > 0)
+            {
+                sb.AppendLine($"{ConfigPath} {{");
+                sb.Append(innerSb);
+                sb.Append("}");
+            }
+
+            sb.Append("}}");
+
+            builder.AddHocon(sb.ToString(), HoconAddMode.Prepend);
+        }
     }
 
     /// <summary>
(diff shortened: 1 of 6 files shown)


Aaronontheweb:
```csharp
 // The base class for the option
 public abstract class DiscoveryOptionBase : IOption
 { }

 // The actual option implementation
 public class ConfigOption : DiscoveryOptionBase
 {
     // Actual option implementation here
     public void Apply(AkkaConfigurationBuilder builder)
     {
         // Modifies Akka.NET configuration either via HOCON or setup class
         builder.AddHocon($"akka.discovery.method = {ConfigPath}", HoconAddMode.Prepend);

         // Rest of configuration goes here
     }
 }

 // Akka.Hosting extension implementation
 public static AkkaConfigurationBuilder WithDiscovery(
     this AkkaConfigurationBuilder builder,
     DiscoveryOptionBase discOption)
 {
     var setup = new DiscoverySetup();

     // gets called here
     discOption.Apply(builder, setup);
 }        
```

I'm not quite seeing the big picture here - how is this an improvement over just prepending config to the front of the HOCON fallback chain? Is the idea here that we instantiate a class that populates the values instead of reading them from a file?

Arkatufus:
Yes, this is to avoid having to provide a HOCON path inside `Akka.Hosting`.

In the example, instead of having the user to remember to use `akka.discovery.config` as the valid string value for the method, they provide the `IOption` implementation instance instead, it is the `IOption` implementation responsibility to provide the HOCON path value value

Arkatufus:
Updated the sample code in the descripition.
Yes, this API is ment for the developers, not the user. And yes, this is just moving the problem from the user to the developer.
The goal is to make it easier for the user to configure Akka.NET via Akka.Hosting, the developer will still need an intimate knowledge of how things are being wired under the hood.

Arkatufus:
It also acts as a "scaffold" for the developer. as in, "if you encounter this kind of HOCON pattern when you create your extension method, this is our standardized way of handling it" so they don't have to re-invent the wheel every time

Arkatufus:
The double abstraction is also intentional, the `IOption` class is a "helper" interface for the developer to use to build their base option class.
The base option class is what the user see, it is the base type for all the options the user can use to configure the extension method and/or `Setup` class

src/Akka.Hosting/IOption.cs
```diff
@@ -0,0 +1,92 @@
...
+    ///         var setup = new DiscoverySetup();
+    /// 
+    ///         // gets called here
+    ///         discOption.Apply(builder, setup);
+    ///     }
+    /// </code>
+    /// </example>
+    public interface IOption
+    {
+        public interface IOption
```
Aaronontheweb, line 65: Did you mean to nest interfaces here?

src/Akka.Hosting/IOption.cs
```diff
@@ -0,0 +1,92 @@
...
+    ///         DiscoveryOptionBase discOption)
+    ///     {
+    ///         var setup = new DiscoverySetup();
+    /// 
+    ///         // gets called here
+    ///         discOption.Apply(builder, setup);
+    ///     }
+    /// </code>
+    /// </example>
+    public interface IOption
```
Aaronontheweb, line 63: I'd rename this to `IHoconOption` so we don't clash with Micrsoft.Extensions.Configuration

Aaronontheweb (commented):
Left some comments

src/Akka.Hosting/IOption.cs
Arkatufus: DOH! I will fix that, thanks

Aaronontheweb:
You know what would really help me here? Refactoring one of our existing configurations to use this - that would help me see, end-to-end, what's going on and what the costs / benefits are.

Arkatufus:
Alright, I'll try to do that.

src/Akka.Cluster.Hosting/SBR/SplitBrainResolverOption.cs
```diff
@@ -84,10 +169,13 @@ public sealed class KeepOldestOption : SplitBrainResolverOption
...
     {
+        public override string ConfigPath => SplitBrainResolverSettings.LeaseMajorityName;
+
         /// <summary>
-        /// The absolute HOCON path to the <see cref="Lease"/> implementation HOCON configuration block
+        /// An class instance that extends <see cref="LeaseOptionBase"/>, used to configure the lease provider used in this
+        /// <see cref="LeaseMajority"/> strategy.
         /// </summary>
-        public string LeaseImplementation { get; set; }
+        public LeaseOptionBase LeaseImplementation { get; set; }
```
Arkatufus, line 178: This is an example of direct use of this standard.
Before, the user need to remember that the lease implementation for `Akka.Coordination.KubernetesApi` is `"akka.coordination.lease.kubernetes"`, now they can actually use intellisense and find `KubernetesLeaseOption` (fake name, not implemented yet) instead.

Aaronontheweb (approved):
LGTM

src/Akka.Cluster.Hosting/SBR/SplitBrainResolverOption.cs
```diff
@@ -38,10 +46,37 @@ public abstract class SplitBrainResolverOption
     /// </summary>
     public sealed class StaticQuorumOption : SplitBrainResolverOption
     {
+        public override string ConfigPath => SplitBrainResolverSettings.StaticQuorumName;
```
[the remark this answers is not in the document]
Arkatufus: Whatever the plugin/module requires, it is the Akka.Hosting module developer duty to know the required HOCON path type to return here.
Historically, we do not have any standard regarding this value, it is absolute in some case, but relative in other.

PR #135, opened 2022-11-02
[Breaking Change] remove default `HoconAddMode` from HOCON builders

### LINKED ISSUE #134: Need to make default `HoconMode` == `Prepend`
Right now the default is `Append`, which means upstream settings (i.e. module defaults) always override downstream calls - which is the opposite of what users intend.

Arkatufus:
Fixes #134

## Changes

* Change AddHocon and AddHoconFile default mode from Append to Prepend

# WARNING
This is a breaking change for all Akka.Hosting extension plugins

## Checklist

* [ ] This change follows the [Akka.NET API Compatibility Guidelines](https://getakka.net/community/contributing/api-changes-compatibility.html).


Arkatufus:
As of the latest discussion with @Aaronontheweb, we will remove the default value instead.

Aaronontheweb (approved):
Breaking change, but this has been a big footgun in Akka.Hosting so far.

Issue #142, opened 2022-11-21
API changes roadmap for 1.0


Arkatufus:
This issue thread is made as a discussion ground on what API needs to be changed/added into `Akka.Hosting` before we're freezing it for 1.0 release. We encourage users to make suggestions in this thread based on their experience in using `Akka.Hosting` so far.

Currently, `Akka.Hosting` still need these API addition/changes:

## Persistence and Cluster Sharding

These API changes are needed to make persistence be compatible with sharding.

- [x] Add API to Akka.Cluster.Hosting to allow user to specify a specific persistence plugin for cluster sharding
- [x] Add API to add a persistence journal/snapshot plugin settings that can be referenced elsewhere. #146
    - Maybe we can expand `AkkaPersistenceJournalBuilder` for journal settings?
    - If so, we need to add `AkkaPersistenceSnapshotBuilder` for snapshot settings
    - Persistence plugins Hosting extensions should not use `Setup` classes, they should inject their HOCON configuration directly.
- [x] Change the current Akka.Persistence.*.Hosting implementation to **not** automatically inject themselves as the default persistence plugins #146
- [x] Change the current Akka.Persistence.*.Hosting API to let their HOCON settings be copied over or renamed to another HOCON block for use with either normal persistence or cluster sharding #146
- [x] Add a generalized Akka.Persistence.Hosting API to specifically set the `akka.persistence.journal.plugin` and `akka.persistence.snapshot-store.plugin` settings. #146

## Remoting

- [x] We will **not** make remote deployment hosting API

## Clustering

- [x] Expand current API with more options to configure the cluster. Current settings candidates: (#149)
    - min-nr-of-members
    - Per role min-nr-of-members
    - app-version
    - Logging related settings
    - Move SBR into this options class (only the new SBR will be supported)
- [x] Expand singleton API to enable lease (#150)
- [x] Expand sharding API to enable lease (#150)

## General
- [ ] Make all options class bindable to Microsoft.Extensions.Configuration

Aaronontheweb:
> Change the current Akka.Persistence.*.Hosting implementation to not automatically inject themselves as the default persistence plugins

I think for most users this is what they'd expect the plugin to do - it's only a handful of power users who want concurrent journals / snapshot stores who'd want the ability to turn this off. We should design the API to support the more common use case which is users running a single journal / snapshot store implementation.

Aaronontheweb:
[...]

PR #146, opened 2022-11-30
Modularize and optionize persistence hosting methods

Arkatufus:
Part of #142

## Changes
- Add `JournalOption`
- Add `SnapshotOption`
- Add the ability to compose event adapters using `JournalOption.Adapters`
- Add the ability to inject journal and snapshot config without making it the default plugin
- Add the Ability to specify which journal and snapshot Sharding uses

Currently supported persistence plugins:
- PostgreSQL
- MS SQLServer

### DIFF
--- a/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
+++ b/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
@@ -45,6 +45,24 @@ namespace Akka.Cluster.Hosting
         public bool RememberEntities { get; set; } = false;
 
         public string Role { get; set; }
+
+        /// <summary>
+        ///     <para>
+        ///         The journal plugin configuration identifier used by persistence mode, eg. "sql-server" or "postgresql"
+        ///     </para>
+        ///     <b>NOTE</b> This setting is only used when <see cref="StateStoreMode"/> is set to
+        ///     <see cref="Akka.Cluster.Sharding.StateStoreMode.Persistence"/>
+        /// </summary>
+        public string JournalPluginId { get; set; } = null;
+
+        /// <summary>
+        ///     <para>
+        ///         The snapshot store plugin configuration identifier used by persistence mode, eg. "sql-server" or "postgresql"
+        ///     </para>
+        ///     <b>NOTE</b> This setting is only used when <see cref="StateStoreMode"/> is set to
+        ///     <see cref="Akka.Cluster.Sharding.StateStoreMode.Persistence"/>
+        /// </summary>
+        public string SnapshotPluginId { get; set; } = null;
     }
 
     public static class AkkaClusterHostingExtensions
@@ -152,11 +170,18 @@ namespace Akka.Cluster.Hosting
         {
             return builder.WithActors(async (system, registry) =>
             {
-                var shardRegion = await ClusterSharding.Get(system).StartAsync(typeName, entityPropsFactory,
-                    ClusterShardingSettings.Create(system)
-                        .WithRole(shardOptions.Role)
-                        .WithRememberEntities(shardOptions.RememberEntities)
-                        .WithStateStoreMode(shardOptions.StateStoreMode), messageExtractor);
+                var settings = ClusterShardingSettings.Create(system)
+                    .WithRole(shardOptions.Role)
+                    .WithRememberEntities(shardOptions.RememberEntities)
+                    .WithStateStoreMode(shardOptions.StateStoreMode);
+
+                if (shardOptions.StateStoreMode == StateStoreMode.Persistence)
+                    settings = settings
+                        .WithJournalPluginId(shardOptions.JournalPluginId)
+                        .WithSnapshotPluginId(shardOptions.SnapshotPluginId);
+                
+                var shardRegion = await ClusterSharding.Get(system)
+                    .StartAsync(typeName, entityPropsFactory, settings, messageExtractor);
                 
                 registry.Register<TKey>(shardRegion);
             });
@@ -190,11 +215,18 @@ namespace Akka.Cluster.Hosting
         {
             return builder.WithActors(async (system, registry) =>
             {
-                var shardRegion = await ClusterSharding.Get(system).StartAsync(typeName, entityPropsFactory,
-                    ClusterShardingSettings.Create(system)
-                        .WithRole(shardOptions.Role)
-                        .WithRememberEntities(shardOptions.RememberEntities)
-                        .WithStateStoreMode(shardOptions.StateStoreMode), extractEntityId, extractShardId);
+                var settings = ClusterShardingSettings.Create(system)
+                    .WithRole(shardOptions.Role)
+                    .WithRememberEntities(shardOptions.RememberEntities)
+                    .WithStateStoreMode(shardOptions.StateStoreMode);
+
+                if (shardOptions.StateStoreMode == StateStoreMode.Persistence)
+                    settings = settings
+                        .WithJournalPluginId(shardOptions.JournalPluginId)
+                        .WithSnapshotPluginId(shardOptions.SnapshotPluginId);
+                
+                var shardRegion = await ClusterSharding.Get(system)
+                    .StartAsync(typeName, entityPropsFactory, settings, extractEntityId, extractShardId);
 
                 registry.Register<TKey>(shardRegion);
             });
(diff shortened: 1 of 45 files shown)


src/Akka.Persistence.Hosting/JournalOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+
+using System;
+using System.Text;
+using Akka.Configuration;
+using Akka.Hosting;
+
+#nullable enable
+namespace Akka.Persistence.Hosting
+{
+    public abstract class JournalOptions
```
Aaronontheweb, line 15: Needs XML-DOC

src/Akka.Persistence.Hosting/JournalOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+        
+        public abstract string Identifier { get; set; }
+        
+        /// <summary>
+        ///     <para>
+        ///         Should corresponding journal table be initialized automatically.
+        ///     </para>
+        ///     <b>Default</b>: false
+        /// </summary>
+        public bool AutoInitialize { get; set; } = false;
```
Aaronontheweb, line 32: Does this also apply to Azure / Redis / MongoDb etc? Or is this setting ignored on platforms that don't support it?

(I'd just make that clear in the XML-DOC)

src/Akka.Persistence.Hosting/JournalOptions.cs
Aaronontheweb: I'd still add this for people who want to add Hosting support to Akka.Persistence plugins - what are some things you need to do to implement this class appropriately / correctly? Doesn't need to be super-detailed, but broad strokes would be good.

src/Akka.Persistence.Hosting/JournalOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+        /// <summary>
+        ///     <para>
+        ///         Default serializer used as manifest serializer when applicable and payload serializer when no
+        ///         specific binding overrides are specified
+        ///     </para>
+        ///     <b>Default</b>: json
+        /// </summary>
+        public string Serializer { get; set; } = "json";
+        
+        public abstract Config DefaultConfig { get; }
```
Aaronontheweb, line 43: Default HOCON for the journal? I'd add `XML-DOC` here too.

src/Akka.Persistence.Hosting/JournalOptions.cs
```diff
@@ -0,0 +1,85 @@
...
+    public abstract class JournalOptions
+    {
+        private readonly bool _isDefault;
+        
+        protected JournalOptions(bool isDefault)
+        {
+            _isDefault = isDefault;
+        }
+        
+        public abstract string Identifier { get; set; }
```
Aaronontheweb, line 24: I'd add XML-DOC here - do implementers need to specify a default identifier?

src/Akka.Persistence.Hosting/SnapshotOptions.cs
```diff
@@ -0,0 +1,79 @@
...
+// -----------------------------------------------------------------------
+
+using System;
+using System.Text;
+using Akka.Configuration;
+using Akka.Hosting;
+
+namespace Akka.Persistence.Hosting
+{
+    public abstract class SnapshotOptions
```
Aaronontheweb, line 14: Same comments I made to `JournalOptions` apply here as well.

src/Akka.Persistence.PostgreSql.Hosting/PostgreSqlJournalOptions.cs
```diff
@@ -0,0 +1,89 @@
...
+
+using System;
+using System.Text;
+using Akka.Configuration;
+using Akka.Persistence.Hosting;
+
+#nullable enable
+namespace Akka.Persistence.PostgreSql.Hosting
+{
+    public sealed class PostgreSqlJournalOptions: SqlJournalOptions
```
Aaronontheweb, line 15: LGTM

Aaronontheweb (commented):
Tentatively - LGTM, but need to see some examples of it being called from the end-user's point of view.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -45,6 +45,24 @@ public sealed class ShardOptions
...
         public string Role { get; set; }
+
+        /// <summary>
+        ///     <para>
+        ///         The journal plugin configuration identifier used by persistence mode, eg. "sql-server" or "postgresql"
+        ///     </para>
+        ///     <b>NOTE</b> This setting is only used when <see cref="StateStoreMode"/> is set to
+        ///     <see cref="Akka.Cluster.Sharding.StateStoreMode.Persistence"/>
+        /// </summary>
+        public string JournalPluginId { get; set; } = null;
```
Aaronontheweb, line 56: LGTM

src/Akka.Persistence.Hosting/SqlSnapshotOptions.cs
```diff
@@ -0,0 +1,74 @@
...
+using Akka.Hosting;
+
+#nullable enable
+namespace Akka.Persistence.Hosting
+{
+    /// <summary>
+    /// Base class for all SQL-based snapshot-store options class. If you're writing an options class for no-SQL or
+    /// other kind of plugins, use <see cref="SnapshotOptions"/> instead.
+    /// </summary>
+    public abstract class SqlSnapshotOptions: SnapshotOptions
```
Aaronontheweb, line 18: LGTM

src/Akka.Persistence.Hosting/JournalOptions.cs
```diff
@@ -0,0 +1,125 @@
...
+        /// The default HOCON configuration for this specific snapshot store configuration identifier, normalized to the
+        /// plugin identifier name
+        /// </summary>
+        public Config DefaultConfig => InternalDefaultConfig.MoveTo(PluginId);
+        
+        /// <summary>
+        /// The journal adapter builder, use this builder to add custom journal
+        /// <see cref="IEventAdapter"/>, <see cref="IWriteEventAdapter"/>, or <see cref="IReadEventAdapter"/>
+        /// </summary>
+        public AkkaPersistenceJournalBuilder Adapters { get; set; } = new ("", null!);
```
Aaronontheweb, line 75: Ah, I see it's possible to add journal-specific adapters here. Brilliant!

src/Akka.Persistence.Hosting/Utils.cs
```diff
@@ -0,0 +1,34 @@
...
+// -----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace Akka.Persistence.Hosting
+{
+    public static class Utils
```
Aaronontheweb, line 14: Make this `internal`

src/Akka.Persistence.PostgreSql.Hosting/StoredAsExtensions.cs
```diff
@@ -0,0 +1,22 @@
...
+//  <copyright file="StoredAsExtensions.cs" company="Akka.NET Project">
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+
+namespace Akka.Persistence.PostgreSql.Hosting
+{
+    public static class StoredAsExtensions
```
Aaronontheweb, line 11: Make this `internal`

Aaronontheweb (commented):
Overall I think this looks great - just left a few questions for you

src/Examples/Akka.Hosting.SqlSharding/Program.cs
```diff
@@ -12,6 +12,53 @@
 
 var builder = WebApplication.CreateBuilder(args);
 
+#if USE_OPTIONS
+builder.Services.AddAkka("MyActorSystem", configurationBuilder =>
+    {
+        var localConn = builder.Configuration.GetConnectionString("sqlServerLocal");
+        var shardingConn = builder.Configuration.GetConnectionString("sqlServerSharding");
+
+        var shardingJournalOptions = new SqlServerJournalOptions(isDefaultPlugin: false)
```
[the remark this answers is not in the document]
Arkatufus: Each options have their own set of adapters, if two persistence journal have the same adapters, each would be need to be set separately, or you can pass the reference to the other, ie. `journalOpt2.Adapters = journalOpt1.Adapters`. Or you can create an adapter builder instance and assign that to both.
The only caveat is that since it is an instance reference, you can't assign it to one and modify the other. We can change this behaviour to do an immutable copy on assignment instead.

src/Examples/Akka.Hosting.SqlSharding/Program.cs
```diff
@@ -12,6 +12,53 @@
...
+        
+        configurationBuilder
+            .WithRemoting("localhost", 8110)
+            .WithClustering(new ClusterOptions()
+            {
+                Roles = new[] { "myRole" },
+                SeedNodes = new[] { Address.Parse("akka.tcp://MyActorSystem@localhost:8110") }
+            })
+            .WithSqlServerPersistence(localConn)
+            .WithSqlServerPersistence(shardingJournalOptions, shardingSnapshotOptions)
```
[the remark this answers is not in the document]
Aaronontheweb: Do we want to log a warning or something if this happens? Or should we not worry about that for now?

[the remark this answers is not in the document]
Arkatufus: We can't really log a warning since we're composing pure HOCON configuration before the `ActorSystem` is being built... I guess we can do a `GetConfig()` and if the ID already exist we throw?

PR #149, opened 2022-12-09
Expand Cluster.Hosting options

Arkatufus:
Part of #142

## Changes
- Expand Cluster.Hosting options
- Add, expand, and clean up XML-DOCs

### DIFF
--- a/src/Akka.Cluster.Hosting.Tests/ClusterOptionsSpec.cs
+++ b/src/Akka.Cluster.Hosting.Tests/ClusterOptionsSpec.cs
@@ -1,0 +1,82 @@
+﻿// -----------------------------------------------------------------------
+//  <copyright file="ClusterOptionsSpec.cs" company="Akka.NET Project">
+//      Copyright (C) 2009-2022 Lightbend Inc. <http://www.lightbend.com>
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using Akka.Actor;
+using Akka.Cluster.Hosting.SBR;
+using Akka.Cluster.SBR;
+using Akka.Configuration;
+using Akka.Hosting;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Akka.Cluster.Hosting.Tests;
+
+public class ClusterOptionsSpec
+{
+    [Fact(DisplayName = "Empty ClusterOptions should contain default HOCON values")]
+    public void EmptyClusterOptionsTest()
+    {
+        var builder = new AkkaConfigurationBuilder(new ServiceCollection(), "")
+            .AddHocon(ConfigurationFactory.FromResource("Akka.Cluster.Configuration.Cluster.conf", typeof(Cluster).Assembly), HoconAddMode.Append)
+            .WithActorRefProvider(ProviderSelection.Cluster.Instance)
+            .BuildClusterHocon(new ClusterOptions());
+        
+        builder.Configuration.HasValue.Should().BeTrue();
+        var settings = new ClusterSettings(builder.Configuration.Value, "");
+
+        settings.Roles.Count.Should().Be(0);
+        settings.AppVersion.CompareTo(Util.AppVersion.Create("assembly-version")).Should().Be(0);
+        settings.MinNrOfMembersOfRole.Count.Should().Be(0);
+        settings.SeedNodes.Count.Should().Be(0);
+        settings.MinNrOfMembers.Should().Be(1);
+        settings.LogInfo.Should().BeTrue();
+        settings.LogInfoVerbose.Should().BeFalse();
+        settings.DowningProviderType.Should().Be(typeof(NoDowning));
+    }
+    
+    [Fact(DisplayName = "ClusterOptions should generate proper HOCON values")]
+    public void ClusterOptionsTest()
+    {
+        var builder = new AkkaConfigurationBuilder(new ServiceCollection(), "")
+            .AddHocon(ConfigurationFactory.FromResource("Akka.Cluster.Configuration.Cluster.conf", typeof(Cluster).Assembly), HoconAddMode.Append)
+            .BuildClusterHocon(new ClusterOptions
+            {
+                Roles = new []{ new Role("front-end"), new Role("back-end", 5)},
+                AppVersion = "1.0.0",
+                MinimumNumberOfMembers = 99,
+                SeedNodes = new [] { Address.Parse("akka.tcp://system@somewhere.com:9999") },
+                LogInfo = false,
+                LogInfoVerbose = true,
+                SplitBrainResolver = new KeepMajorityOption
+                {
+                    Role = "back-end"
+                }
+            });
+        
+        builder.Configuration.HasValue.Should().BeTrue();
+        var settings = new ClusterSettings(builder.Configuration.Value, "");
+
+        settings.Roles.Should().BeEquivalentTo("front-end", "back-end");
+        
+        settings.MinNrOfMembersOfRole.Count.Should().Be(1);
+        settings.MinNrOfMembersOfRole.ContainsKey("back-end").Should().BeTrue();
+        settings.MinNrOfMembersOfRole["back-end"].Should().Be(5);
+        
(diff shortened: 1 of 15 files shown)


src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -13,38 +13,170 @@
...
+    ///     The set of options for enabling Akka.Cluster support.
     /// </summary>
     public sealed class ClusterOptions
     {
         /// <summary>
-        /// The akka.cluster.roles values.
+        ///     The akka.cluster.roles and akka.cluster.role values.
         /// </summary>
-        public string[] Roles { get; set; }
+        public Role[]? Roles { get; set; }
```
Aaronontheweb, line 52: Don't like the `Role` type - adds more friction to the configuration surface in order to expose a feature barely anybody uses. Users can configure this in HOCON if they really want it - please remove it from the UI surface area entirely and go back to using simple strings.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -13,38 +13,170 @@
...
 
         /// <summary>
-        /// If populated, the akka.cluster.seed-nodes that will be used.
+        ///     If populated, the akka.cluster.seed-nodes that will be used.
+        /// </summary>
+        public Address[]? SeedNodes { get; set; }
```
Aaronontheweb, line 57: So here's a goal worth considering with the `ClusterOptions` class - should this be parseable from an `appsettings.json` file / Environment Variable? If so, maybe we should support `string` here instead and just handle the parsing on the back-end. I'm leaning towards doing that for 1.0.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -13,38 +13,170 @@
...
+        /// </summary>
+        public string? Role { get; set; }
+        
+        /// <summary>
+        /// When handing over to a new oldest node this <see cref="TerminationMessage"/> is sent to the singleton actor
+        /// to tell it to finish its work, close resources, and stop. The hand-over to the new oldest node
+        /// is completed when the singleton actor is terminated. Note that <see cref="PoisonPill"/> is a
+        /// perfectly fine <see cref="TerminationMessage"/> if you only need to stop the actor.
+        /// </summary>
+        public object? TerminationMessage { get; set; }
```
Aaronontheweb, line 157: LGTM

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -13,38 +13,170 @@
...
     }
 
     public sealed class ShardOptions
     {
+        /// <summary>
```
Aaronontheweb, line 162: All of the `ShardOptions` changes LGTM

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Aaronontheweb: Yeah, let's change this to a `string` type

Aaronontheweb (changes requested):
Need to lean more towards ease of use / low friction even if it means not exposing certain settings in Akka.Hosting. Goal of this library is low friction, not completeness.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Aaronontheweb: Here's what we have to do in the current stable version of Akka.Hosting:

```csharp
var akkaSection = context.Configuration.GetSection("Akka");

// maps to environment variable Akka__ClusterIp
var hostName = akkaSection.GetValue<string>("ClusterIp", "localhost");

// maps to environment variable Akka__ClusterPort
var port = akkaSection.GetValue<int>("ClusterPort", 0);

var seeds = akkaSection.GetValue<string[]>("ClusterSeeds", new[] { "akka.tcp://SqlSharding@localhost:7918" })
    .Select(Address.Parse)
    .ToArray();
```

It would be amazing if we could just do this instead (and if we decide to go down this route, we should write a unit test that looks like this):

**`appsettings.json`**

```json
{
    "Logging": {
        "LogLevel": {
            "Default": "Debug",
            "System": "Information",
            "Microsoft": "Information"
        }
    },
    "Akka": {
        "RemoteOptions":{
            "Hostname": "localhost",
            "Port": 8081,
            "PublicHostname": "localhost",
            "PublicPort": 8081
        },
        "ClusterOptions": {
            "SeedNodes": "akka.tcp://ClusterSystem@localhost:8081",
            "Roles": ["backend"]
        }
    }
}
```

**`program.cs`**

```csharp
var akkaSection = context.Configuration.GetSection("Akka");
var clusterConfig = akkaSection.Get<ClusterOptions>(); // ideally, this should work
```

We've not really stated this as an explicit goal for the project before, but in the context of our "pit of success" mission for this library I think it's probably worth doing before we ship 1.0. What do you think?

src/Akka.Cluster.Hosting.Tests/ClusterOptionsSpec.cs
```diff
@@ -0,0 +1,152 @@
...
+    },
+    ""KeepMajorityOption"": {
+      ""Role"" : ""back-end""
+    }
+  }
+}";
+        using var stream = new MemoryStream(Encoding.UTF8.GetBytes(json));
+        var jsonConfig = new ConfigurationBuilder().AddJsonStream(stream).Build();
+        
+        var clusterOptions = jsonConfig.GetSection("Akka:ClusterOptions").Get<ClusterOptions>();
```
Arkatufus, line 125: Binding directly from json provided configuration is totally doable now.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -13,38 +13,151 @@
...
+        ///     <para>
+        ///         Minimum required number of members before the leader changes member status
+        ///         of 'Joining' members to 'Up'. Typically used together with
+        ///         <see cref="Cluster.RegisterOnMemberUp"/> to defer some action, such as starting actors,
+        ///         until the cluster has reached a certain size.
+        ///     </para>
+        ///     <b>Default:</b> 1
+        /// </summary>
+        public int? MinimumNumberOfMembers { get; set; }
```
Arkatufus, line 49: Would this be a better compromise to the `Role` class?
We're not forcing the user to provide any values and still keeps the class as simple as possible.

Arkatufus (commented):
Self review for the last change request

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Arkatufus: ah, wrong line...

src/Akka.Cluster.Hosting.Tests/ClusterOptionsSpec.cs
Aaronontheweb: Looks great

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -13,38 +13,151 @@
...
-        public string[] Roles { get; set; }
+        public string[]? Roles { get; set; }
 
         /// <summary>
-        /// If populated, the akka.cluster.seed-nodes that will be used.
+        ///     Optional cluster role check to consider if a specific cluster role have enough
+        ///     members to be considered to be up. The default value is 1 node per role. 
         /// </summary>
-        public Address[] SeedNodes { get; set; }
+        public Dictionary<string, int>? MinimumNumberOfMembersPerRole { get; set; }
```
Arkatufus, line 33: This is the line I ment.
Would this be a better compromise to the Role class?
We're not forcing the user to provide any values and still keeps the class as simple as possible.

Aaronontheweb, line 33: Yep, I think that's fine

Aaronontheweb (approved):
LGTM - let's auto-merge this once you resolve your merge conflict

PR #150, opened 2022-12-09
Add lease support to singleton and sharding Akka.Hosting extensions

Arkatufus:
Part of #142

## Changes

* Move LeaseOptionBase from Akka.Cluster.Hosting to Akka.Hosting
* Add Lease support to Akka.Cluster.Sharding Hosting extensions option
* Add Lease support to Akka.Cluster.Singleton Hosting extensions option

### DIFF
--- a/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
+++ b/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
@@ -11,7 +11,9 @@ using Akka.Cluster.Tools.Client;
 using Akka.Cluster.Tools.PublishSubscribe;
 using Akka.Cluster.Tools.Singleton;
 using Akka.Configuration;
+using Akka.Coordination;
 using Akka.Hosting;
+using Akka.Hosting.Coordination;
 
 namespace Akka.Cluster.Hosting
 {
@@ -36,6 +38,17 @@ namespace Akka.Cluster.Hosting
         public int? BufferSize { get; set; } = null;
         public string Role { get; set; }
         public object TerminationMessage { get; set; }
+        
+        /// <summary>
+        /// An class instance that extends <see cref="LeaseOptionBase"/>, used to configure the lease provider used in this
+        /// cluster singleton.
+        /// </summary>
+        public LeaseOptionBase? LeaseImplementation { get; set; }
+        
+        /// <summary>
+        /// The interval between retries for acquiring the lease
+        /// </summary>
+        public TimeSpan? LeaseRetryInterval { get; set; }
     }
 
     public sealed class ShardOptions
@@ -63,6 +76,17 @@ namespace Akka.Cluster.Hosting
         ///     <see cref="Akka.Cluster.Sharding.StateStoreMode.Persistence"/>
         /// </summary>
         public string SnapshotPluginId { get; set; } = null;
+        
+        /// <summary>
+        /// An class instance that extends <see cref="LeaseOptionBase"/>, used to configure the lease provider used in this
+        /// sharding region.
+        /// </summary>
+        public LeaseOptionBase? LeaseImplementation { get; set; }
+        
+        /// <summary>
+        /// The interval between retries for acquiring the lease
+        /// </summary>
+        public TimeSpan? LeaseRetryInterval { get; set; }
     }
 
     public static class AkkaClusterHostingExtensions
@@ -175,6 +199,13 @@ namespace Akka.Cluster.Hosting
                     .WithRememberEntities(shardOptions.RememberEntities)
                     .WithStateStoreMode(shardOptions.StateStoreMode);
 
+                if (shardOptions.LeaseImplementation is { })
+                {
+                    var retry = shardOptions.LeaseRetryInterval ?? TimeSpan.FromSeconds(5);
+                    settings = settings
+                        .WithLeaseSettings(new LeaseUsageSettings(shardOptions.LeaseImplementation.ConfigPath, retry));
+                }
+
                 if (shardOptions.StateStoreMode == StateStoreMode.Persistence)
                     settings = settings
                         .WithJournalPluginId(shardOptions.JournalPluginId)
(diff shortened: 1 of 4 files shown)


src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -36,6 +38,17 @@ public sealed class ClusterSingletonOptions
         public int? BufferSize { get; set; } = null;
         public string Role { get; set; }
         public object TerminationMessage { get; set; }
+        
+        /// <summary>
+        /// An class instance that extends <see cref="LeaseOptionBase"/>, used to configure the lease provider used in this
+        /// cluster singleton.
+        /// </summary>
+        public LeaseOptionBase? LeaseImplementation { get; set; }
```
Aaronontheweb, line 46: I think this is kind of ugly - reason being is that all of the other `*Options` classes are POCOs but `LeaseOptionsBase` combines effects (`Apply(AkkaConfigurationBuilder builder, Setup setup = null);`) and data - it'd be cleaner if we could keep the implementations pure. How could we go about doing this?

Aaronontheweb (commented):
Have some questions

src/Akka.Hosting/Coordination/LeaseOptionBase.cs
```diff
@@ -0,0 +1,18 @@
...
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//  </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using Akka.Actor.Setup;
+
+namespace Akka.Hosting.Coordination
+{
+    public abstract class LeaseOptionBase : IHoconOption
```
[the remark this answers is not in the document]
Arkatufus: Historically, `Lease` is part of `Akka.Coordination` which is outside of `Akka.Cluster`. But since we don't have `Akka.Coordination.Hosting`, the closest we can do is to move it to `Akka.Hosting`.

I guess this is just future proving the feature so that it is not tied to clustering in general.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Arkatufus: `LeaseOptionBase` is the exact same way we're doing it in SBR, the same identical option can be used for both SBR and sharding at the same time.

src/Akka.Hosting/Coordination/LeaseOptionBase.cs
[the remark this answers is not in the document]
Aaronontheweb: But is there anything outside of clustering that would need a lease? Even though they're separate libraries, aren't the clustering components really the only consumers of this?

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Aaronontheweb: That part I get, but it's the combination of state + effects in one of the configuration classes that I'm iffy on.

Arkatufus: It is practically (for now) a placeholder to point the sharding plugin to the actual configuration path being used by the implementing `Lease` class itself.
For example, the implementation in kubernetes is here: https://github.com/akkadotnet/Akka.Management/blob/51638f2acb51efaea64415fc5dd7df214e49f63b/src/coordination/kubernetes/Akka.Coordination.KubernetesApi/KubernetesLeaseOption.cs

Arkatufus: It does have the ability to modify/inject more code into `AkkaConfigurationBuilder` if needed, that part is a future proofing feature. `IHoconOption` itself is __not__ intended for end user, it is more of a template an `Akka.Hosting` developer had to use to implement a very specific HOCON pattern. The documentation is here: https://github.com/akkadotnet/Akka.Hosting/blob/6502ac3ac58490b2514befd4877e6b69a20aeb93/src/Akka.Hosting/IHoconOption.cs#L12-L62

Aaronontheweb: Ok, we'll leave that be then

Arkatufus: Prettyfying it:

# IHoconOption

Standardized interface template for a common HOCON configuration pattern where a configuration takes a HOCON config path and the config path contains a class FQCN property and other settings that are needed by said class.

The pattern looks like this:
```hocon
# This HOCON property references to a config block below
akka.discovery.method = akka.discovery.config

akka.discovery.config {
    class = "Akka.Discovery.Config.ConfigServiceDiscovery, Akka.Discovery"
    # other options goes here
}
```
## Example
Example implementation for the pattern described above
```csharp
// The base class for the option (templating)
public abstract class DiscoveryOptionBase : IOption
{ }

// The actual option implementation
public class ConfigOption : DiscoveryOptionBase
{
    // Actual option implementation here
     public void Apply(AkkaConfigurationBuilder builder)
     {
         // Modifies Akka.NET configuration either via HOCON or setup class
         builder.AddHocon($"akka.discovery.method = {ConfigPath}", HoconAddMode.Prepend);

         // Rest of configuration goes here
     }
}

 // Akka.Hosting extension implementation
 public static AkkaConfigurationBuilder WithDiscovery(
     this AkkaConfigurationBuilder builder,
     DiscoveryOptionBase discOption)
 {
     var setup = new DiscoverySetup();

     // gets called here
     discOption.Apply(builder, setup);
 }
```

Arkatufus: The user isn't supposed to see all this horribleness, what they needed to do is to set
`new ClusterSingletonOptions { LeaseImplementation = KubernetesLeaseOption.Instance };`

Aaronontheweb (approved):
LGTM

PR #157, opened 2022-12-19
Expand Cluster.Hosting options to also use the new JournalOptions and SnapshotOptions

Arkatufus:
## Changes

Expand Cluster.Hosting options to also use the new JournalOptions and SnapshotOptions as optional properties


Aaronontheweb (approved):
LGTM

PR #158, opened 2022-12-19
Add IConfiguration to HOCON adapter

Arkatufus:
## Changes
Adds a variation to `AddHocon` that can convert Microsoft.Extensions.Configuration `IConfiguration`  into HOCON `Config` instance and adds it to the
ActorSystem being configured.

### NOTES
* All variable name are automatically converted to lower case.
* All "." (period) in the `IConfiguration` key will be treated as a HOCON object key separator
* For environment variable configuration provider:
  * "__" (double underline) will be converted to "." (period).
  * "_" (single underline) will be converted to "-" (dash).
  * If all keys are composed of integer parseable keys, the whole object is treated as an array

### Example
JSON configuration:
```json
{
   "akka.cluster": {
       "roles": [ "front-end", "back-end" ],
       "min-nr-of-members": 3,
       "log-info": true
   }
}
```
and environment variables:
```powershell
AKKA__CLUSTER__ROLES__0=front-end
AKKA__CLUSTER__ROLES__1=back-end
AKKA__CLUSTER__MIN_NR_OF_MEMBERS=3
AKKA__CLUSTER__LOG_INFO=true
```  
is equivalent to HOCON configuration of:
```HOCON
 akka {
     cluster {
         roles: [ front-end, back-end ]
         min-nr-of-members: 3
         log-info: true
     }
 }
```

### DIFF
--- a/src/Akka.Hosting.Tests/Akka.Hosting.Tests.csproj
+++ b/src/Akka.Hosting.Tests/Akka.Hosting.Tests.csproj
@@ -4,8 +4,8 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="$(MicrosoftExtensionsVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="$(MicrosoftExtensionsVersion)" />
     <PackageReference Include="Akka.TestKit.Xunit2" Version="$(AkkaVersion)" />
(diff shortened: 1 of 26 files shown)


src/Akka.Hosting.Tests/Akka.Hosting.Tests.csproj
```diff
@@ -4,6 +4,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="7.0.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.EnvironmentVariables" Version="7.0.0" />
```
Aaronontheweb, line 8: Should probably use the `$(MicrosoftExtensionsVersion)` substitution here

Arkatufus: Done

Aaronontheweb (approved):
LGTM

PR #160, opened 2022-12-20
Add IConfiguration bindable RemoteOptions to Remote.Hosting

Arkatufus:
## Changes

- Add IConfiguration bindable RemoteOptions to Remote.Hosting
- XML-DOCs
- Unit tests

### DIFF
--- a/src/Akka.Remote.Hosting/Akka.Remote.Hosting.csproj
+++ b/src/Akka.Remote.Hosting/Akka.Remote.Hosting.csproj
@@ -3,6 +3,8 @@
     <TargetFramework>$(LibraryFramework)</TargetFramework>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <Description>Akka.Remote Microsoft.Extensions.Hosting support.</Description>
+    <LangVersion>9.0</LangVersion>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
   <ItemGroup>
--- a/src/Akka.Remote.Hosting/RemoteOptions.cs
+++ b/src/Akka.Remote.Hosting/RemoteOptions.cs
@@ -4,17 +4,39 @@
 //  </copyright>
 // -----------------------------------------------------------------------
 
-using System;
+using System.Net;
 using System.Text;
-using Akka.Configuration;
 
 namespace Akka.Remote.Hosting
 {
     public sealed class RemoteOptions
     {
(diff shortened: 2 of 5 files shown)


src/Akka.Remote.Hosting/RemoteOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+using Akka.Configuration;
+
+namespace Akka.Remote.Hosting
+{
+    public sealed class RemoteOptions
+    {
+        public string? HostName { get; set; }
+        public int? Port { get; set; }
+        public string? PublicHostName { get; set; }
+        public int? PublicPort { get; set; }
```
Aaronontheweb, line 18: I would add an XML-DOC comment saying that this param is optional and virtually no one needs this - only in instances where you need to map a public port to a different private port is this necessary.

src/Akka.Remote.Hosting/RemoteOptions.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Text;
+using Akka.Configuration;
+
+namespace Akka.Remote.Hosting
+{
+    public sealed class RemoteOptions
+    {
+        public string? HostName { get; set; }
+        public int? Port { get; set; }
+        public string? PublicHostName { get; set; }
```
Aaronontheweb, line 17: I would mention that public hostname will default to `HostName` when not populated

Basically re-use what's included in the comments in the reference hocon for Akka.Remote

src/Akka.Remote.Hosting/Akka.Remote.Hosting.csproj
```diff
@@ -3,6 +3,8 @@
     <TargetFramework>$(LibraryFramework)</TargetFramework>
     <PackageReadmeFile>README.md</PackageReadmeFile>
     <Description>Akka.Remote Microsoft.Extensions.Hosting support.</Description>
+    <LangVersion>9.0</LangVersion>
+    <Nullable>enable</Nullable>
```
Aaronontheweb, line 7: Can we set this via `Directory.Build.props` and just ensure that this is followed everywhere? Might as well start Akka.Hosting off on the right foot.

Aaronontheweb (changes requested):
API looks good, just have minor recommendations

Aaronontheweb (approved):

PR #161, opened 2022-12-20
Make JournalOptions and SnapshotOptions IConfiguration bindable

Arkatufus:
## Changes
- Make JournalOptions and SnapshotOptions IConfiguration bindable
- Make journal adapters bindable via IConfiguration (advanced options)


Aaronontheweb (changes requested):
I'm going to mark this as "don't merge" for now - we'd ultimately prefer users to configure adapters and other "live" objects through C#. Adding a second way of doing this through `IConfiguration`, while nice, is something we should probably only do if there's demand for it. Since this change appears to be non-breaking (or at least, could be implemented that way) I think we can save it for the future.

Arkatufus:
Fixed, removed the adapter part.

Aaronontheweb (approved):
LGTM

PR #166, opened 2022-12-23
[Cluster.Hosting] Fix nullable warnings and modernize codes

Arkatufus:


src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -824,7 +824,7 @@ internal static Config CreateReceptionistConfig(string name, string? role)
                 .Append(root).Append("name:").AppendLine(name.ToHocon());
             
             if(!string.IsNullOrEmpty(role))
-                sb.Append(root).Append("role:").AppendLine(role.ToHocon());
+                sb.Append(root).Append("role:").AppendLine(role!.ToHocon());
```
Aaronontheweb, line 827: LGTM - `ToHocon` has `null` handling built into it, IIRC? @Arkatufus

Aaronontheweb (commented):
Left a question

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Arkatufus: Its compiler/roslyn error, its already checked for null by the line before it.

Aaronontheweb (approved):
LGTM

PR #178, opened 2022-12-29
Fix Hosting.TestKit logging problem

### LINKED ISSUE #177: Akka.Hosting.TestKit: just have `ConfigureLogging` "do the right thing" by default and automatically wire up `ITestOutput`

Each time I write a test I have to remember how to pipe the MSFT Logging, ILoggingAdapter, and the xUnit `ITestOutputHelper` together and it's very cumbersome.

base implementation of `protected virtual void ConfigureLogging(ILoggingBuilder builder){}` should just do this for me unless I specify otherwise. Maybe even make the `ITestOutputHelper` a required constructor parameter (breaking change but no one else is using this yet)

Arkatufus:
Fixes #177

The way the logger works inside Akka.Hosting and Akka.Hosting.TestKit is different. TestKit logger needs to actually listen to the event bus because it is wired manually.

## Changes

- Make LoggerFactoryLogger methods virtual so it can be overriden
- Inharit LoggerFactoryLogger class in TestKit specialized logger and override the Receive method


Aaronontheweb (approved):
LGTM

PR #180, opened 2023-01-03
Better string to HOCON literal conversion, accounting for all illegal characters

Arkatufus:
## Changes
Better string to HOCON literal conversion:
- accounts for all illegal characters
- support for triple quote
- support for null


Aaronontheweb (approved):
LGTM

PR #207, opened 2023-02-07
Fix string.ToHocon() backslash escape character compatibility

Arkatufus:
## Changes
* Add escape character compatibility to string.ToHocon()
* Remove triple quote support, not needed because we now use escape characters


Aaronontheweb (approved):
LGTM

PR #217, opened 2023-02-16
Update PR validation script

Arkatufus:


.github/workflows/pr_validation.yml
```diff
@@ -21,10 +21,12 @@ on:
     branches:
       - master
       - dev
+      - v1.4
```
Aaronontheweb: `v1.*` (for future-proofing)

Aaronontheweb (changes requested):

.github/workflows/pr_validation.yml
Arkatufus: Good idea, didn't know that this is an option

Aaronontheweb (approved):

Aaronontheweb:
@Arkatufus also going to need to backport this to v1.4 :p

.github/workflows/pr_validation.yml
```diff
@@ -51,10 +51,8 @@ jobs:
             .nuke/temp
             ~/.nuget/packages
           key: ${{ runner.os }}-${{ hashFiles('**/global.json', '**/*.csproj') }}
-      - name: Run './build.cmd All'
-        run: ./build.cmd All
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Run './build.cmd RunTests'
```
Aaronontheweb, line 54: So a comment here - on each PR we should run:

- Build
- Tests
- NuGet packaging

The `All` target shouldn't attempt to do a NuGet __RELEASE__ without the credentials being passed in.

We'll need to do another PR to fix this.

Arkatufus: Got it, i'll do a fix

PR #224, opened 2023-02-16
Add new Cluster.Sharding RememberEntitiesStore option

Arkatufus:


src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -157,17 +157,38 @@ public sealed class ClusterSingletonOptions
...
+        ///     <see cref="Akka.Cluster.Sharding.StateStoreMode.DData"/>, this controls how the remembered entities
+        ///     and shards are stored.
+        ///     </para>
+        ///
+        ///     <para>
+        ///     Possible values are <see cref="Akka.Cluster.Sharding.RememberEntitiesStore.Eventsourced"/> and
+        ///     <see cref="Akka.Cluster.Sharding.RememberEntitiesStore.DData"/>
+        ///     </para> 
+        /// 
+        ///     Default is <see cref="Akka.Cluster.Sharding.RememberEntitiesStore.DData"/> for backwards compatibility.
```
Aaronontheweb: I think the default here should be none / ignored by default: https://getakka.net/community/whats-new/akkadotnet-v1.5-upgrade-advisories.html - so I'd set this to `null` I suppose and handle that accordingly.

Aaronontheweb: I think your code looks ok - it might just be the XML-DOC comment that's wrong.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -239,6 +260,55 @@ public sealed class ShardOptions
...
+        {
+            var sb = new StringBuilder();
+
+            if (Role is { })
+                sb.AppendLine($"role = {Role.ToHocon()}");
+            
+            if(RememberEntities is { })
+                sb.AppendLine($"remember-entities = {RememberEntities.ToHocon()}");
+            
+            if(RememberEntitiesStore is { })
```
Aaronontheweb, line 279: So if the value is not populated, don't set it in HOCON - that looks correct to me for backwards compat IMHO

Aaronontheweb (changes requested):
I think the XML-DOC on the `RememberEntitiesStore` property might be wrong.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Arkatufus: Well, the default is `RememberEntitiesStore.DData`, its just not set inside the options class, it is set from the default configuration

Arkatufus: https://github.com/akkadotnet/akka.net/blob/c9ccc25207b5a4cfa963a5a23f96c0676fbbef10/src/contrib/cluster/Akka.Cluster.Sharding/reference.conf#L26-L29

Aaronontheweb: Sigh, it's wrong in both places then. Let's fix it here and then submit a PR to do it in Akka.NET as well.

Arkatufus: So if the user did not set the `RememberEntitiesStore` property, the default value would be taken from the fallback default configuration, which is `"ddata"`

Aaronontheweb: Good catch 🔍

Aaronontheweb (approved):

PR #236, opened 2023-02-21
[fix] GitHubAction offline

eaba:


Aaronontheweb (approved):
LGTM - need to backport to v1.4 and our `dotnet new` templates.

PR #249, opened 2023-03-02
Set version range for Akka.Hosting Msft.Ext to [3,)

Aaronontheweb:
No need for us to target .NET 7.0 APIs so tightly.


Aaronontheweb:
(we'll see if the CI/CD system bears that out)

PR #264, opened 2023-03-15
Add `GetAync` method to `IRequiredActor<TActor>` to resolve issues where actor is not available upon injection (i.e. `BackgroundService`s)

### LINKED ISSUE #208: BackgroundService DI of IRequiredActor for reference to an Actor fails with 'No actor registered'
**Version Information**
Akka.Hosting 1.0.2

**Describe the bug**
Creating a `Microsoft.Extensions.Hosting.BackgroundService` with a DI fails to inject a IRequiredActor interface with the error of `Akka.Hosting.MissingActorRegistryEntryException: 'No actor registered for key <ActorName>'`

**To Reproduce**
The simplest code reproduction of this error is
```
using Akka.Actor;
using Akka.Hosting;

var builder = WebApplication.CreateBuilder(args);

builder.Services.AddHostedService<ThingRunningInTheBackground>();

builder.Services.AddAkka("mysystem", configurationBuilder =>
{
    configurationBuilder.WithActors((system, registry, resolver) =>
    {
        var leadActor = system.ActorOf(Props.Create(() => new LeadActor()));

        registry.TryRegister<LeadActor>(leadActor);
    });
});

var app = builder.Build();
app.MapGet("/", () => "Hello World!");
app.Run();

public class LeadActor : ReceiveActor
{
    public LeadActor()
    {
        Receive<string>(Console.Write);
    }
}

public class ThingRunningInTheBackground : BackgroundService
{
    private readonly IActorRef _actor;

    public ThingRunningInTheBackground(IRequiredActor<LeadActor> actor)
    {
        _actor = actor.ActorRef;
    }
    protected override Task ExecuteAsync(CancellationToken stoppingToken)
    {
        throw new NotImplementedException();
    }
}
```
**Environment**
Windows 10
Net 7.0

**Discord Discussions**
Chatting on discord about this problem @Aaronontheweb 

***What the problem is***
> I'm not sure how to work around this problem inside IRequiredActor<T> just yet
> but the issue is that all of your actors are started in another IHostedService managed by [Akka.NET](http://akka.net/)
> and that hasn't had a chance to run and populate the ActorRegistry by the time your BackgroundService starts
> so until I come up with a way to force IRequiredActor<T> to do this behind the scenes (probably by blocking, unless there's an await-able way of doing this)

***The current solution to the approach***
> so the solution to this is to have the ActorRegistry injected into your BackgroundService and then call ActorRegistry.GetAsync<T> on it
> that call will return a Task<IActorRef> you can await on
> and the task will be completed as soon as the actor gets added to the registry
> which should only be a on the order of 100s of milliseconds

Tweet by Aaron asking about some approachs
https://twitter.com/Aaronontheweb/status/1623064967724298251

Taking the above simple reproduction this below code works noting the injection of `ActorRegistry` into the constructor and using the `_actorRegistry.GetAsync` to get the actor reference.
```
using Akka.Actor;
using Akka.Hosting;

var builder = WebApplication.CreateBuilder(args);
builder.Services.AddHostedService<ThingRunningInTheBackgroundUsingRegistry>();

builder.Services.AddAkka("mysystem", configurationBuilder =>
{
    configurationBuilder.WithActors((system, registry, resolver) =>
    {
        var leadActor = system.ActorOf(Props.Create(() => new LeadActor()));

        registry.TryRegister<LeadActor>(leadActor);
    });
});

var app = builder.Build();
app.MapGet("/", () => "Hello World!");
app.Run();

public class LeadActor : ReceiveActor
{
    public LeadActor()
    {
        Receive<string>(Console.Write);
    }
}

public class ThingRunningInTheBackgroundUsingRegistry : BackgroundService
{
    private IActorRef? _actorReference;
    private readonly ActorRegistry _actorRegistry;

    public ThingRunningInTheBackgroundUsingRegistry(ActorRegistry actorRegistry)
    {
        _actorRegistry = actorRegistry;
    }
    protected override async Task ExecuteAsync(CancellationToken stoppingToken)
    {
        _actorReference = await _actorRegistry.GetAsync<LeadActor>(stoppingToken);

        _actorReference.Tell("Message");
    }
}
```

Aaronontheweb:
## Changes

Fixes #208 - working on fix to ensure that other `BackgroundService` / `IHostedService` implementations can still depend on `IRequiredActor<TActor>` without getting blown up by MSFT.EXT.DI while it waits for the `AkkaService` to start.

## Checklist

* [x] Design discussion issue #208

### DIFF
--- a/src/Akka.Hosting.Tests/Akka.Hosting.Tests.csproj
+++ b/src/Akka.Hosting.Tests/Akka.Hosting.Tests.csproj
@@ -17,6 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Akka.Hosting.TestKit\Akka.Hosting.TestKit.csproj" />
     <ProjectReference Include="..\Akka.Hosting\Akka.Hosting.csproj" />
   </ItemGroup>
 
--- a/src/Akka.Hosting.Tests/Bugfix208Specs.cs
+++ b/src/Akka.Hosting.Tests/Bugfix208Specs.cs
@@ -1,0 +1,87 @@
+﻿using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Actor.Dsl;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Xunit;
+
+namespace Akka.Hosting.Tests;
+
+public class Bugfix208Specs : TestKit.TestKit
+{
+    private class MyTestActor : ReceiveActor
+    {
+        public record SetData(string Data);
+
+        public record GetData();
+        
+        private string _data = string.Empty;
+        
+        public MyTestActor()
+        {
+            Receive<SetData>(s =>
+            {
+                _data = s.Data;
+            });
+            
+            Receive<GetData>(g =>
+            {
+                Sender.Tell(_data);
+            });
+        }
+    }
+    
+    private class TestActorKey{}
+    
+    private class MyBackgroundService : BackgroundService
+    {
+        private readonly IActorRef _testActor;
+
+        public MyBackgroundService(IRequiredActor<TestActorKey> requiredActor)
+        {
+            _testActor = requiredActor.ActorRef;
+        }
+
+        protected override Task ExecuteAsync(CancellationToken stoppingToken)
+        {
+            _testActor.Tell("BackgroundService started");
+            return Task.CompletedTask;
+        }
+    }
+
+    protected override void ConfigureServices(HostBuilderContext context, IServiceCollection services)
+    {
+        services.AddHostedService<MyBackgroundService>();
+        base.ConfigureServices(context, services);
+    }
+
+    protected override void ConfigureAkka(AkkaConfigurationBuilder builder, IServiceProvider provider)
+    {
+        builder.WithActors((system, registry, arg3) =>
+        {
+            registry.Register<TestActorKey>(system.ActorOf(Props.Create(() => new MyTestActor()), "test-actor"));
+        });
+    }
+
+    /// <summary>
+    /// Reproduction for https://github.com/akkadotnet/Akka.Hosting/issues/208
+    /// </summary>
+    [Fact]
+    public async Task ShouldStartHostedServiceThatDependsOnActor()
+    {
+        // arrange
+        var testActorRef = ActorRegistry.Get<TestActorKey>();
+
+        // act
+
+        // assert
+        await AwaitAssertAsync(async () =>
+        {
+            var r = await testActorRef.Ask<string>(new MyTestActor.GetData(), TimeSpan.FromMilliseconds(100));
+            r.Should().Be("BackgroundService started");
+        });
+    }
+}
--- a/src/Akka.Hosting.Tests/Bugfix208Specs.cs
+++ b/src/Akka.Hosting.Tests/Bugfix208Specs.cs
@@ -38,17 +38,17 @@ public class Bugfix208Specs : TestKit.TestKit
     
     private class MyBackgroundService : BackgroundService
     {
-        private readonly IActorRef _testActor;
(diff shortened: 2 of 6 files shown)


src/Akka.Hosting.Tests/Akka.Hosting.Tests.csproj
```diff
@@ -17,6 +17,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\Akka.Hosting.TestKit\Akka.Hosting.TestKit.csproj" />
```
Aaronontheweb, line 20: Needed to add the Hosting.TestKit in order to test the `BackgroundService` issue with #208

src/Akka.Hosting.Tests/Bugfix208Specs.cs
```diff
@@ -0,0 +1,108 @@
...
+    public async Task ShouldStartHostedServiceThatDependsOnActor()
+    {
+        // arrange
+        var testActorRef = ActorRegistry.Get<TestActorKey>();
+
+        // act
+
+        // assert
+        
+        // workaround for https://github.com/akkadotnet/Akka.Hosting/issues/265
```
Aaronontheweb, line 82: This is the reproduction for #265

src/Akka.Hosting.Tests/Bugfix208Specs.cs
```diff
@@ -0,0 +1,108 @@
...
+    
+    private class TestActorKey{}
+    
+    private class MyBackgroundService : BackgroundService
+    {
+        private readonly IRequiredActor<TestActorKey> _testActor;
+
+        public MyBackgroundService(IRequiredActor<TestActorKey> requiredActor)
+        {
+           _testActor = requiredActor;
```
Aaronontheweb, line 45: So we tried to resolve the `IRequiredActor<TActor>.ActorRef` here, like users did in #208, but it won't work - if we block the constructor of a `BackgroundService` here it'll stop the `AkkaService` and another services from starting. It's too fragile of a fix for end-users, so instead we're going with throwing a clearer exception when an `ActorRegistry` miss occurs and offer an additional API on the `IRequiredActor<TActor>` interface to address it.

src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -23,6 +24,28 @@ public interface IRequiredActor<TActor>
...
+        /// <summary>
+        /// When calling from inside another <see cref="IHostedService"/>, actor registrations may not be
+        /// available at startup (due to Akka.NET itself being started asynchronously in another hosted service).
+        ///
+        /// Instead - you should call the GetAsync method to wait for that actor to be populated into the <see cref="ActorRegistry"/>
+        /// by the AkkaService at startup.
+        /// </summary>
+        /// <param name="cancellationToken">Optional cancellation token.</param>
+        /// <returns>A Task that will return the <see cref="IActorRef"/> using the given key.</returns>
+        Task<IActorRef> GetAsync(CancellationToken cancellationToken = default);
```
Aaronontheweb, line 37: This is the correct fix - offer a method on the `IRequiredActor<TActor>` interface that uses the `ActorRegistry.GetAsync` under the covers. Users will need to call this if they fall under the #208 edge case.

src/Akka.Hosting/ActorRegistry.cs
```diff
@@ -31,13 +54,53 @@ public interface IRequiredActor<TActor>
...
+                    return _internalRef;
+
+                // attempt 2 - synchronously check the registry (fast path)
+                if (_registry.TryGet<TActor>(out _internalRef))
+                {
+                    return _internalRef;
+                }
+
+
+                throw new MissingActorRegistryEntryException(
```
Aaronontheweb, line 71: Try our best to give the users a clear signal as to why this failed and what they should do instead. We can't really solve the problem for them and the normal, sync `ActorRef` property works for 99.9% of cases so we don't want to remove that just to support an edge case.

Aaronontheweb (commented):
Reviewed my own code

Arkatufus (approved):
LGTM

PR #274, opened 2023-04-05
Change default persistence journal and snapshot store serializer to null

Arkatufus:
## Changes
* Change default journal serializer to null
* Change default snapshot store serializer to null


Aaronontheweb (commented):
What are the effects of this change?

Arkatufus:
Basically turning off the persistence default serializer setting and use the default Akka object serializer instead.

Aaronontheweb:
> Basically turning off the persistence default serializer setting and use the default Akka object serializer instead.

Sounds good to me. We'll need to merge in v1.5.2 from Akka.NET into here to "double tap" this change.

PR #284, opened 2023-04-24
Add transaction isolation level support

Arkatufus:
## Changes

* Add `IsolationLevel` properties to SqlJournalOptions and SqlSnapshotOptions base class
* Add `IsolationLevel` properties to Akka.Persistence.SqlServer.Hosting
* Add `IsolationLevel` properties to Akka.Persistence.PostgreSql.Hosting

### DIFF
--- a/src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApprovePersistence.verified.txt
+++ b/src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApprovePersistence.verified.txt
@@ -23,6 +23,11 @@ namespace Akka.Persistence.Hosting
         public Akka.Persistence.Hosting.AkkaPersistenceJournalBuilder AddWriteEventAdapter<TAdapter>(string eventAdapterName, System.Collections.Generic.IEnumerable<System.Type> boundTypes)
             where TAdapter : Akka.Persistence.Journal.IWriteEventAdapter { }
     }
+    public static class Extensions
+    {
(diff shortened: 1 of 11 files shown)


src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApprovePersistence.verified.txt
```diff
@@ -64,19 +69,23 @@ namespace Akka.Persistence.Hosting
         public string ConnectionString { get; set; }
         public System.TimeSpan ConnectionTimeout { get; set; }
         public abstract string MetadataTableName { get; set; }
+        public abstract System.Data.IsolationLevel ReadIsolationLevel { get; set; }
```
Aaronontheweb, line 72: Shouldn't this only be in the Sql-specific plugins?

Aaronontheweb (commented):
Left a question

src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApprovePersistence.verified.txt
Arkatufus: Yes, this change is inside `SqlJournalOptions` class

Aaronontheweb: Ah - I didn't realize those types were in the base Akka.Persistence.Hosting package.

PR #290, opened 2023-05-01
Add sharding `passivate-idle-entity-after` property

Arkatufus:
## Changes

* Add `ShardOptions.PassivateIdleEntityAfter` option property


Aaronontheweb (approved):
LGTM

PR #293, opened 2023-05-04
Add MAUI support

### LINKED ISSUE #289: MAUI support for Akka.Hosting (disable `IHostedService`)

I spent some time chasing down a [.NET MAUI](https://dotnet.microsoft.com/en-us/apps/maui) user's issue getting `IRequiredActor<TKey>` to work correctly inside their application. Their `ActorRegistry` settings were correct and their code correctly binded Akka.NET to the `MauiAppBuilder` - the `IHost` equivalent for MAUI:

```csharp
public static MauiAppBuilder Akka(this MauiAppBuilder builder)
{
    builder.Services.AddAkka("MyActorSystem", akkaConfigurationBuilder =>
    {
        akkaConfigurationBuilder.PipeActor();
    });

    return builder;
}
```

Yet the `IRequiredActor<TKey>` injection doesn't work and throws an actor resolution error upon trying, even with Akka.Hosting v1.5.4+ - why is that?

This is because MAUI doesn't support `IHostedServices`: https://github.com/dotnet/maui/issues/2244

Assuming that the rest of the core Akka DLL can even run on MAUI (i.e. there are no major AOT or other issues that would cause it to run not on the client), one solution we could try is configuring Akka.Hosting to run and launch all actors without the `IHostedService`. I think it might be possible to remove the `IHostedService` altogether without very little downside for regular users, possibly:

* We can still shut the `ActorSystem` down using the `IApplicationLifetime` and continuation `Task`s without needing the `IHostedService.StopAsync` method;
* We already start the `ActorSystem` prior to the `IHostedService.StartAsync` method;
* We already have a separate data structure for keeping track of what needs to be launched at startup - we could, in theory, just execute that as-is in the foreground with no code changes;
* We already blow up the foreground process and kill it if there are problems starting any of the actors or `Startup` tasks - we'd just be doing that with one less layer of indirection under this arrangement.

To implement this, we'd need an integration test that demonstrates Akka.NET failing to boot up under MAUI now - and we'd need to get that test to turn green by removing the `AkkaHostedService`.

Alternatively, rather than removing the `AkkaHostedService` we could add an extension method that starts up Akka.NET without the `IHostedService` - or we could add an explicit MAUI build target and using compiler directives that avoid the use of the `IHostedService` in MAUI only. Those would lower the risk of Akka.Hosting having adverse side-effects for existing users.

**Additional context**

Arkatufus:
Fixes #289

## Changes
* Add partial MAUI support

## NOTES
* We will never fully support MAUI officially, i.e. we will not bundle/include MAUI SDK into Akka.Hosting
* This will be a discussion thread for MAUI support, we will strip the MAUI sample code on the actual merge-able PR (non-compilable by CI/CD, missing MAUI SDK).


src/Akka.Hosting/AkkaHostedService.cs
```diff
@@ -17,11 +17,11 @@ internal sealed class AkkaHostedService : IHostedService
         private CoordinatedShutdown? _coordinatedShutdown; // grab a reference to CoordinatedShutdown early
         private readonly IServiceProvider _serviceProvider;
         private readonly AkkaConfigurationBuilder _configurationBuilder;
-        private readonly IHostApplicationLifetime _hostApplicationLifetime;
+        private readonly IHostApplicationLifetime? _hostApplicationLifetime;
         private readonly ILogger<AkkaHostedService> _logger;
 
         public AkkaHostedService(AkkaConfigurationBuilder configurationBuilder, IServiceProvider serviceProvider,
-            ILogger<AkkaHostedService> logger, IHostApplicationLifetime applicationLifetime)
+            ILogger<AkkaHostedService> logger, IHostApplicationLifetime? applicationLifetime)
```
Arkatufus, line 24: MAUI does not support `IHostApplicationLife`, we need to pass in `null` and disable termination hook.

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -63,8 +64,23 @@ public static IServiceCollection AddAkka(this IServiceCollection services, strin
             // registers the hosted services and begins execution
             b.Bind();
             
-            // start the IHostedService which will run Akka.NET
-            services.AddHostedService<AkkaHostedService>();
+            if (Util.IsRunningInMaui)
+            {
+                services.AddSingleton<AkkaHostedService>(provider =>
```
Arkatufus: MAUI does not support hosting, hosted services had to be registered as a singleton service.

src/Akka.Hosting/MauiSupport.cs
```diff
@@ -0,0 +1,30 @@
...
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Akka.Hosting
+{
+    public static class AkkaMauiSupport
+    {
+        public static async Task StartAkka(IServiceProvider provider, CancellationToken token = default)
```
Arkatufus: Akka hosted service have to be started manually by users, `AkkaHostedService` is internal, we have to use these wrapper methods to let user call the service `StartAsync()` method

src/Akka.Hosting/Util.cs
```diff
@@ -14,6 +15,17 @@ namespace Akka.Hosting
 {
     public static class Util
     {
+        // HACK: MAUI runtime detection
+        private static bool? _runningInMaui;
+        internal static bool IsRunningInMaui
+        {
+            get
+            {
+                _runningInMaui ??= AppDomain.CurrentDomain.GetAssemblies().Any(asm => asm.GetName().Name.StartsWith("Microsoft.Maui"));
```
Arkatufus, line 24: MAUI runtime detection

src/Examples/Akka.Hosting.Maui/MauiProgram.cs
```diff
@@ -0,0 +1,45 @@
...
+            builder.Services
+                .AddAkka("TestSys", config =>
+                {
+                    config.WithActors((system, registry) =>
+                    {
+                        var echo = system.ActorOf(ClickActor.Props);
+                        registry.Register<ClickActor>(echo);
+                    });
+                })
+                .AddTransient<IMauiInitializeService, AkkaInitializer>()
```
Arkatufus: Register a `IMauiInitializeService` to start Akka hosted service.
This is the only way to start a service automatically during start-up in MAUI.

src/Examples/Akka.Hosting.Maui/MauiProgram.cs
```diff
@@ -0,0 +1,45 @@
...
+            var app = builder.Build();
+            return app;
+        }
+    }
+    
+    public class AkkaInitializer: IMauiInitializeService
+    {
+        public void Initialize(IServiceProvider services)
+        {
+            AkkaMauiSupport.StartAkka(services).GetAwaiter().GetResult();
```
Arkatufus: Akka hosted service starter, we call the wrapped service here.

Aaronontheweb (approved):
Adds MAUI detection, but all of the actual code for making Akka.Hosting workable in MAUI will be located in its own repository here: https://github.com/akkadotnet/Akka.Hosting.Maui

src/Akka.Hosting/Util.cs
MacKey-255: Problem detecting Microsoft.Maui in an Android project using Microsoft.Maui.Essentials.
Error using AddAkka:
```
System.PlatformNotSupportedException: Due to https://github.com/dotnet/maui/issues/2244, normal Akka.Hosting.AddAkka method will not work.Instead, you need to install Akka.Hosting.Maui and use the AddAkkaMaui extension method instead.
   at Akka.Hosting.AkkaHostingExtensions.AddAkka[AkkaHostedService](IServiceCollection services, String actorSystemName, Action`2 builder)
   at Akka.Hosting.AkkaHostingExtensions.AddAkka(IServiceCollection services, String actorSystemName, Action`2 builder)
```
   
Error using AddAkkaMaui:
```
System.IO.FileNotFoundException: Could not load file or assembly 'Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null' or one of its dependencies.
File name: 'Microsoft.Maui, Version=1.0.0.0, Culture=neutral, PublicKeyToken=null'
   at MyApp.Mobile.Services.AkkaService.AddAkkaMobile(IServiceCollection services, IHostEnvironment environment) in 
   ...
```

Aaronontheweb: Is this the library you're looking for? https://github.com/akkadotnet/Akka.Hosting.Maui/

MacKey-255: Noup :/ Check [it](https://github.com/akkadotnet/Akka.Hosting/issues/592) out pls

PR #318, opened 2023-05-26
Fix `PassivateIdleEntityAfter` to not override HOCON settings

### LINKED ISSUE #303: Akka.Cluster.Hosting: `ShardOptions.PassivateIdleEntityAfter` overwrites custom `akka.cluster.sharding.passivate-idle-entity-after` setting in HOCON
**Version Information**
Version of Akka.NET? v1.5.6
Which Akka.NET Modules? Akka.Cluster.Hosting

**Describe the bug**

Not 100% if this is a bug or not yet (going off an end-user's report), but if I set

```hocon
akka.cluster.sharding.passivate-idle-entity-after = 24h #24 hours
```

In a separate `.hocon` file added via the `AddHoconFile` method (with prepending et al done correctly), but if I start any `ShardRegion` using the default `ShardOptions` (without specifying `ShardOptions.PassivateIdleEntityAfter`) I get the following logs:

image

This indicates that we're still using the default 2 minutes setting, and I believe that's because of our default value for `ShardOptions` here:

https://github.com/akkadotnet/Akka.Hosting/blob/9e08da29910ef9c395b7de6c758724e0b8ad78ef/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs#L288

**To Reproduce**

1. Create `ActorSystem` with `akka.cluster.sharding.passivate-idle-entity-after = 24h`
2. Start `ShardRegion` using Akka.Hosting;
3. Check to see if the passivation timeout is 2 minutes instead of 24 hours.

**Additional context**

We should probably make `null` the default value here so that way the HOCON values will be used if not user-supplied value is defined.

Arkatufus:
Fixes #303

## Changes
* Make `PassivateIdleEntityAfter` default value to null
* Add `ShouldPassivateIdleEntities` property to disable idle entity passivation

### DIFF
--- a/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
+++ b/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
@@ -280,19 +280,24 @@ namespace Akka.Cluster.Hosting
         public ShardingDDataOptions DistributedData { get; } = new();
 
         /// <summary>
-        /// Set this to a time duration to have sharding passivate entities when they have not
-        /// received any message in this length of time. Set to <c>null</c> to disable.
-        /// It is always disabled if <see cref="RememberEntities"/> is enabled.
+        /// Set this to false to disable idle entity passivation. When set to <c>false</c>,
+        /// will always override <see cref="PassivateIdleEntityAfter"/>
         /// </summary>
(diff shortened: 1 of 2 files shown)


Aaronontheweb (approved):
LGTM

PR #329, opened 2023-06-16
Return non-zero exit code when `CoordinatedShutdown` indicates problems

### LINKED ISSUE #328: Akka.Hosting: return non-zero exit code when `CoordinatedShutdown` is invoked due to error'd reasons

In order to enable process supervision on platforms like Windows Services to properly restart an Akka.Hosting `/ `IHost`-powered application, we need to return a non-zero exit code when the application exits improperly.

In many cases, things are shutdown for normal reasons - however, due to the [`CoordinatedShutdown.Reason` property](https://getakka.net/api/Akka.Actor.CoordinatedShutdown.Reason.html) we can generally determine cases where the process may not have shut down cleanly:

* [`CoordinatedShutdown.ClusterJoinUnsuccessfulReason`](https://getakka.net/api/Akka.Actor.CoordinatedShutdown.ClusterJoinUnsuccessfulReason.html)
* [`CoordinatedShutdown.ClusterDowningReason`](https://getakka.net/api/Akka.Actor.CoordinatedShutdown.ClusterDowningReason.html)

We should return a non-zero exit code when one of those two types is returned during the method below:

https://github.com/akkadotnet/Akka.Hosting/blob/95c2a2465f8104deeb13a4396466c2b183445bf6/src/Akka.Hosting/AkkaHostedService.cs#L47-L51

It might be worth considering extending the `CoordinatedShutdown.Reason` class to include an `IsError` property that would indicate whether or not we should log a non-zero exit code and incorporate this functionality directly into `CoordinatedShutdown` instead, but I think that might be a bit too prescriptive.

Aaronontheweb:
Fixes #328

## Changes

add a non-zero exit code in the event of an unclean shutdown

## Checklist

* [x] Design discussion issue #328

### DIFF
--- a/src/Akka.Hosting/AkkaHostedService.cs
+++ b/src/Akka.Hosting/AkkaHostedService.cs
@@ -47,6 +47,19 @@ namespace Akka.Hosting
                 async Task TerminationHook()
                 {
                     await ActorSystem.WhenTerminated.ConfigureAwait(false);
+                    
+                    /*
+                     * Set a non-zero exit code in the event that we get a known, confirmed unclean shutdown
+                     * from the ActorSystem / CoordinatedShutdown
+                     */
+                    switch (CoordinatedShutdown.ShutdownReason)
+                    {
+                        case CoordinatedShutdown.ClusterDowningReason _:
+                        case CoordinatedShutdown.ClusterLeavingReason _:
+                            Environment.ExitCode = -1;
+                            break;
+                    }
+
                     HostApplicationLifetime?.StopApplication();
                 }
 


Aaronontheweb:
Tried automatic a test for this using [Tmds.ExecFunction](https://github.com/tmds/Tmds.ExecFunction) but ran into issues with the .NET runtime not being able to find assemblies (probably due to all of the dependencies we're bringing in), so I created a separate EXE and ran this:

image

Validates a non-zero exit code for bad reasons, unmodified (usually `0`) for everything else.

src/Akka.Hosting/AkkaHostedService.cs
```diff
@@ -47,6 +47,19 @@ public virtual async Task StartAsync(CancellationToken cancellationToken)
...
                 {
                     await ActorSystem.WhenTerminated.ConfigureAwait(false);
+                    
+                    /*
+                     * Set a non-zero exit code in the event that we get a known, confirmed unclean shutdown
+                     * from the ActorSystem / CoordinatedShutdown
+                     */
+                    switch (CoordinatedShutdown.ShutdownReason)
+                    {
+                        case CoordinatedShutdown.ClusterDowningReason _:
```
Aaronontheweb, line 57: We know for sure there are errors with these two shutdown reasons - all of the others are uncertain. So only return a non-zero exit code when we are 100% certain.

Arkatufus: We might want to differentiate these exit codes in the future.

Arkatufus: should we add `ClusterJoinUnsuccessfulReason`?

Arkatufus (approved):
LGTM, just nitpicking a bit.

src/Akka.Hosting/AkkaHostedService.cs
Aaronontheweb: > We might want to differentiate these exit codes in the future.

I think that's a bit of YAGNI - in this instance we just need to return something non-zero so the Windows Service supervision system knows the process didn't exit cleanly.

Aaronontheweb: The logs will make it clear what actually happened

PR #337, opened 2023-07-12
Fix missing ClusterClient default config when clustering was not used.

Arkatufus:
Reported in Discord, ClusterClient does not have the correct default HOCON configuration when `.WithClustering()` was not called.

## Changes

* Add default config in extension method


Aaronontheweb (approved):
LGTM

PR #363, opened 2023-08-28
Fix IConfiguration to HOCON adapter

### LINKED ISSUE #351: IConfiguration To HOCON Adapter always converts all keys to lower-case
**Version Information**
Version of Akka.NET? 1.5,12
Which Akka.NET Modules?

**Describe the bug**
In HOCON I can configure Akka.Net to use the Hyperion serializer like this:

````hocon
  actor {
    serializers {
      hyperion = "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
    }
    serialization-bindings {
      "System.Object" = hyperion
    }
  }
````

I tried to move my HOCON configuration to ASP.NET IConfiguration (in appsettings.json).
Fist I had a problem with the dot character in "System.Object" because the HOCON adapter treats all keys as potentially dotted paths. I found out I can quote the key like this in my appsettings.json (this should probably be documented):

````json
{
  "akka": {
    "actor": {
      "serializers": {
        "hyperion": "Akka.Serialization.HyperionSerializer, Akka.Serialization.Hyperion"
      },
      "serialization-bindings": {
        "\"System.Object\"": "hyperion"
      }
    }
  }
}
````

I have still the problem that the HOCON adapter converts all keys to lowercase (see ExpandKey in ConfigurationHoconAdapter.cs). I get the following warning at startup:

````
[WARNING][08.11.2023 08:29:24.190Z][Thread 0001][ActorSystem(EnsoBackend)] The type name for message/serializer binding 'hyperion' did not resolve to an actual Type: 'system.object'
````

and if I log the HCON config with `"log-config-on-start": true` I can see the problem:

image

### PR DESCRIPTION
**Pull request by Arkatufus**
Fixes #351

## Changes

* Default behaviour change, only configurations pulled from the environment variables are automatically converted to lower case
* Update documentation

### DIFF
--- a/src/Akka.Hosting/Configuration/EnvironmentVariablesConfigurationHoconAdapter.cs
+++ b/src/Akka.Hosting/Configuration/EnvironmentVariablesConfigurationHoconAdapter.cs
@@ -1,0 +1,41 @@
+﻿// -----------------------------------------------------------------------
+//   <copyright file="EnvironmentVariablesConfigurationHoconAdapter.cs" company="Akka.NET Project">
+//      Copyright (C) 2013-2023 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//   </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Reflection;
+using Microsoft.Extensions.Configuration.EnvironmentVariables;
+
+namespace Akka.Hosting.Configuration;
+
+internal class EnvironmentVariablesConfigurationHoconAdapter: EnvironmentVariablesConfigurationProvider
+{
+    public static EnvironmentVariablesConfigurationHoconAdapter Create(EnvironmentVariablesConfigurationProvider parent)
+    {
+        var propertyInfo = typeof(EnvironmentVariablesConfigurationProvider).GetProperty("Data", BindingFlags.Instance | BindingFlags.NonPublic);
+        if (propertyInfo is null)
+            throw new Exception($"{nameof(EnvironmentVariablesConfigurationProvider)} source code has changed: Could not find protected property 'Data'");
+        
+        var prefix = (Dictionary<string, string>)propertyInfo.GetValue(parent);
+        return new EnvironmentVariablesConfigurationHoconAdapter(prefix);
+    }
+    
+    private EnvironmentVariablesConfigurationHoconAdapter(Dictionary<string, string> data)
+    {
+        foreach (var kvp in data)
+        {
+            // Sanitize configuration brought in from environment variables, 
+            // "__" are already converted to ":" by the environment configuration provider.
+            var newKey = kvp.Key.ToLowerInvariant().Replace("_", "-");
+            Data[newKey] = kvp.Value;
+        }
+    }
+
+    public override void Load()
+    {
+        throw new NotImplementedException();
+    }
+}
--- a/src/Akka.Hosting/Configuration/HoconConfigurationRoot.cs
+++ b/src/Akka.Hosting/Configuration/HoconConfigurationRoot.cs
@@ -1,0 +1,74 @@
+﻿// -----------------------------------------------------------------------
+//   <copyright file="HoconConfigurationRoot.cs" company="Akka.NET Project">
+//      Copyright (C) 2013-2022 .NET Foundation <https://github.com/akkadotnet/akka.net>
+//   </copyright>
+// -----------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.EnvironmentVariables;
+using Microsoft.Extensions.Primitives;
+
+namespace Akka.Hosting.Configuration;
+
+internal class HoconConfigurationRoot: IConfigurationRoot
+{
+    private readonly IList<IConfigurationProvider> _providers;
+
+    public HoconConfigurationRoot(IEnumerable<IConfigurationProvider> providers)
+    {
+        _providers = providers.Select(provider => 
+            provider is not EnvironmentVariablesConfigurationProvider envProvider ? 
+                provider : EnvironmentVariablesConfigurationHoconAdapter.Create(envProvider)).ToList();
+    }
+
+    public IConfigurationSection GetSection(string key) => new ConfigurationSection(this, key);
+
+    public HoconConfigurationSection GetHoconSection(string key) => new (this, key);
+
+    public IEnumerable<IConfigurationSection> GetChildren()
+    {
+        return _providers
+            .Aggregate(Enumerable.Empty<string>(),
+                (seed, source) => source.GetChildKeys(seed, null))
+            .Distinct(StringComparer.OrdinalIgnoreCase)
+            .Select(GetSection).ToList();
+    }
+
+    public IEnumerable<HoconConfigurationSection> GetHoconChildren()
+    {
+        return _providers
+            .Aggregate(Enumerable.Empty<string>(),
+                (seed, source) => source.GetChildKeys(seed, null))
+            .Distinct(StringComparer.OrdinalIgnoreCase)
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

File: src/Akka.Hosting/Configuration/EnvironmentVariablesConfigurationHoconAdapter.cs [hunk 1]
```diff
@@ -0,0 +1,41 @@
...
+using System.Reflection;
+using Microsoft.Extensions.Configuration.EnvironmentVariables;
+
+namespace Akka.Hosting.Configuration;
+
+internal class EnvironmentVariablesConfigurationHoconAdapter: EnvironmentVariablesConfigurationProvider
+{
+    public static EnvironmentVariablesConfigurationHoconAdapter Create(EnvironmentVariablesConfigurationProvider parent)
+    {
+        var propertyInfo = typeof(EnvironmentVariablesConfigurationProvider).GetProperty("Data", BindingFlags.Instance | BindingFlags.NonPublic);
```
Arkatufus, line 18: We have to do some reflection magic here, we can not re-create the `IConfigurationRoot` because some `IConfigurationProvider` can not be run twice (such as the `StreamConfigurationProvider`), so we have to do a raw copy of the existing environment variable provider by reflecting its internal data.

src/Akka.Hosting/Configuration/EnvironmentVariablesConfigurationHoconAdapter.cs
```diff
@@ -0,0 +1,41 @@
...
+        return new EnvironmentVariablesConfigurationHoconAdapter(prefix);
+    }
+    
+    private EnvironmentVariablesConfigurationHoconAdapter(Dictionary<string, string> data)
+    {
+        foreach (var kvp in data)
+        {
+            // Sanitize configuration brought in from environment variables, 
+            // "__" are already converted to ":" by the environment configuration provider.
+            var newKey = kvp.Key.ToLowerInvariant().Replace("_", "-");
```
Arkatufus, line 32: We do all of the key normalization/transformation here. From this point on, all of the data inside this provider wrapper is the valid HOCON data that we needed

src/Akka.Hosting/Configuration/HoconConfigurationRoot.cs
```diff
@@ -0,0 +1,74 @@
...
+
+internal class HoconConfigurationRoot: IConfigurationRoot
+{
+    private readonly IList<IConfigurationProvider> _providers;
+
+    public HoconConfigurationRoot(IEnumerable<IConfigurationProvider> providers)
+    {
+        _providers = providers.Select(provider => 
+            provider is not EnvironmentVariablesConfigurationProvider envProvider ? 
+                provider : EnvironmentVariablesConfigurationHoconAdapter.Create(envProvider)).ToList();
```
Arkatufus, line 24: We're doing a provider list copy here, replacing/wrapping the official `EnvironmentVariableConfigurationProvider` with our own

src/Akka.Hosting/Configuration/HoconConfigurationRoot.cs
```diff
@@ -0,0 +1,74 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.EnvironmentVariables;
+using Microsoft.Extensions.Primitives;
+
+namespace Akka.Hosting.Configuration;
+
+internal class HoconConfigurationRoot: IConfigurationRoot
```
Arkatufus, line 16: This is just a wrapper over the official `IConfigurationRoot` implementation, it replaces any `EnvironmentVariablesConfigurationProvider` with our own implementation.

src/Akka.Hosting/Configuration/HoconConfigurationRoot.cs
```diff
@@ -0,0 +1,74 @@
...
+    public IEnumerable<HoconConfigurationSection> GetHoconChildren()
+    {
+        return _providers
+            .Aggregate(Enumerable.Empty<string>(),
+                (seed, source) => source.GetChildKeys(seed, null))
+            .Distinct(StringComparer.OrdinalIgnoreCase)
+            .Select(GetHoconSection).ToList();
+    }
+    
+    public IChangeToken GetReloadToken() => throw new NotImplementedException();
```
Arkatufus, line 49: All write support methods are not implemented, this class in an internal only class and will only be used for read when we're generating the HOCON `Config` instance.

src/Akka.Hosting/Configuration/EnvironmentVariablesConfigurationHoconAdapter.cs
```diff
@@ -0,0 +1,41 @@
...
+            Data[newKey] = kvp.Value;
+        }
+    }
+
+    public override void Load()
+    {
+        throw new NotImplementedException();
```
Arkatufus, line 39: We're not loading anything, all the data are derived from the original provider.
This class should never be used as an actual provider, its just a helper class to convert `IConfiguration` to HOCON `Config`

src/Akka.Hosting/Configuration/HoconConfigurationSection.cs
```diff
@@ -0,0 +1,83 @@
...
+            .Distinct(StringComparer.OrdinalIgnoreCase)
+            .Select(key => Root.GetHoconSection(ConfigurationPath.Combine(Path, key)));
+    }
+
+    public IChangeToken GetReloadToken() => Root.GetReloadToken();
+
+    public string? this[string key]
+    {
+        get => Root[ConfigurationPath.Combine(Path, key)];
+        set => throw new NotImplementedException();
```
Arkatufus, line 68: All write support methods are not implemented, this class in an internal only class and will only be used for read when we're generating the HOCON `Config` instance.

src/Akka.Hosting/Configuration/HoconConfigurationSection.cs
```diff
@@ -0,0 +1,83 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Primitives;
+
+namespace Akka.Hosting.Configuration;
+
+internal class HoconConfigurationSection: IConfigurationSection
```
Arkatufus, line 16: This class wraps the official `ConfigurationSection` class and automatically inspects the providers to see if they need to be converted to `EnvironmentVariablesConfigurationHoconAdapter`.

src/Akka.Hosting/Configuration/HoconConfigurationSection.cs
```diff
@@ -0,0 +1,83 @@
...
+{
+    public static HoconConfigurationSection Create(IConfigurationSection section)
+    {
+        if (section is HoconConfigurationSection hoconSection)
+            return hoconSection;
+
+        if (section is not ConfigurationSection configSection)
+            throw new Exception($"{nameof(section)} argument could not be cast to {nameof(ConfigurationSection)}");
+
+        var fieldInfo = typeof(ConfigurationSection).GetField("_root", BindingFlags.Instance | BindingFlags.NonPublic);
```
Arkatufus, line 26: We need to do some reflection magic here to retrieve the original root configuration providers. This needs to be done because these roots are the actual origin of all configuration values, any environment variable provider need to be replaced with our own provider.

Arkatufus (commented):
Self review

Arkatufus:
Solution rejected due to use of reflection magic

mjalil:
Hi @Arkatufus,
I also not pleasant with this solution, as it uses reflection. And has some methods that return `NotImplementedException`, it reminds me a smell of Interface Segregation problem. Another issue IMHO is that the code is not easy to understand, simple and needs comments.

@Aaronontheweb 
My first proposal is detecting strings in double quotes and lowercase others, as I see in the HOCON, only strings enclosed in double quotes are case-sensitive, such as `"System.Object" = hyperion` in `serialization-bindings`

PR #365, opened 2023-08-29
Fix IConfiguration adapter

Arkatufus:


README.md
```diff
@@ -392,6 +410,96 @@ var host = new HostBuilder()
     });
 ```
 
+### Advanced Usage
```
Aaronontheweb: Rename to "special characters and case sensitivity"

src/Akka.Hosting/Configuration/ConfigurationHoconAdapter.cs
```diff
@@ -15,20 +15,20 @@ namespace Akka.Hosting.Configuration
 {
     public static class ConfigurationHoconAdapter
     {
-        public static Config ToHocon(this IConfiguration config)
+        public static Config ToHocon(this IConfiguration config, bool normalizeKeys = true)
```
Aaronontheweb, line 18: LGTM

Aaronontheweb (changes requested):
We need to add a spec to test this change (handling of case sensitivity)

Arkatufus:
@Aaronontheweb the spec for case sensitivity is there already

Aaronontheweb (approved):
LGTM

PR #423, opened 2024-01-08
Add method to configure IHostBuilder in TestKit

Arkatufus:
## Changes

Add `ConfigureHostBuilder()` method to `Akka.Hosting.TestKit` so user can configure the `IHostBuilder` instance directly. Needed so that user can plug in other MS.EXT.Hosting extensions such as Castle Windsor that needed to access `IHostBuilder` to configure itself.


Aaronontheweb (approved):
LGTM

PR #460, opened 2024-06-04
Add log filtering integration

Arkatufus:


README.md
```diff
@@ -636,3 +636,57 @@ To set up the `Microsoft.Extensions.Logging` log filtering, you will need to edi
 ```
 
 [Back to top](#akkahosting)
+
+<a id="filtering-logs"></a>
+## Filtering Logs In Akka.NET
+
+In Akka.NET 1.5.21, we introduced [log filtering for log messages based on the LogSource or the content of a log message](https://getakka.net/articles/utilities/logging.html#filtering-log-messages). Depending on your coding style, you can use this feature in Akka.Hosting in several ways.
+
+1. Using the `WithLogFilter()` `AkkaConfigurationBuilder` extension method.
```
Aaronontheweb: LGTM

README.md
```diff
@@ -636,3 +636,57 @@ To set up the `Microsoft.Extensions.Logging` log filtering, you will need to edi
...
+   {
+       configurationBuilder
+           .WithLogFilter(filterBuilder =>
+           {
+               filterBuilder.ExcludeMessageContaining("Test");
+           });
+   });
+   ```
+
+2. Using The `LoggerConfigBuilder.WithLogFilter()` method.
```
Aaronontheweb: LGTM

src/Akka.Hosting.Tests/HostingExtensionsSpec.cs
```diff
@@ -32,4 +34,33 @@ public void WithActorAskTimeoutInfiniteTest()
         builder.Configuration.HasValue.Should().BeTrue();
         builder.Configuration.Value.GetString("akka.actor.ask-timeout").Should().Be("infinite");
     }
+    
+    [Fact(DisplayName = "WithLogFilter should inject LogFilterSetup")]
+    public void WithLogFilterSetupTest()
+    {
+        var builder = new AkkaConfigurationBuilder(new ServiceCollection(), "fake")
```
Aaronontheweb: LGTM

src/Akka.Hosting/LoggerConfigBuilder.cs
```diff
@@ -96,6 +98,13 @@ public LoggerConfigBuilder ClearLoggers()
             return this;
         }
 
+        public LoggerConfigBuilder WithLogFilter(Action<LogFilterBuilder> filterBuilder)
```
Aaronontheweb, line 101: One thing I'm not so sure about - what happens if someone calls this and `ConfigureLogging` (or whatever it's called?) - can't we end up with one config overriding the other?

Aaronontheweb (commented):
One question about the two different ways of doing this, but otherwise looks good

src/Akka.Hosting/LoggerConfigBuilder.cs
Arkatufus: Yes, the two implementation inside `ConfigureLoggers(builder)` and the extension method `AkkaConfigurationBuilder.WithLogFilter()` is exclusive to each other. The first being added would win because of how we retrieve `Setup` classes in `ActorSystem.Settings`

Aaronontheweb: LGTM - we decided to stick with just the `LogConfigBuilder`.

README.md
```diff
@@ -642,22 +642,7 @@ To set up the `Microsoft.Extensions.Logging` log filtering, you will need to edi
...
 
-1. Using the `WithLogFilter()` `AkkaConfigurationBuilder` extension method.
-
-   The `WithLogFilter()` extension method lets you set up the `LogFilterBuilder` 
-
-   ```csharp
-   builder.Services.AddAkka("MyActorSystem", configurationBuilder =>
-   {
-       configurationBuilder
-           .WithLogFilter(filterBuilder =>
```
Aaronontheweb: LGTM - glad we consolidated this.

Aaronontheweb (approved):
LGTM (once API approvals are in)

PR #478, opened 2024-07-25
Add ClusterClientDiscovery Hosting extension support

Arkatufus:
## Changes

* [x] Add `ClusterClientDiscoveryOptions` class
* [x] Add new `ClusterClient` extensions methods `WithClusterClientDiscovery` that allows users to use `ClusterClientDiscovery`
* [x] Add XML-DOC
* [x] Add unit tests
* [ ] Update main documentation


src/Akka.Cluster.Hosting/Akka.Cluster.Hosting.csproj
```diff
@@ -13,5 +13,9 @@
   <ItemGroup>
     <PackageReference Include="Akka.Cluster.Sharding" Version="$(AkkaVersion)" />
     <PackageReference Include="Akka.DistributedData" Version="$(AkkaVersion)" />
+    <PackageReference Include="Akka.Management" Version="$(AkkaManagementVersion)" />
```
Aaronontheweb: We should not be doing this - we didn't need to bring all of these dependencies in for Cluster Bootstrap, certainly don't want to bring them in for something as trivial as `ClusterClient`.

Aaronontheweb (changes requested):
Can't reference any part of Akka.Management in this project. That will create a dependency cycle.

src/Akka.Cluster.Hosting/Akka.Cluster.Hosting.csproj
Arkatufus: done

src/Akka.Cluster.Hosting/ClusterClientDiscoveryOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+    ///         The discovery sub-system that will be used to discover cluster client contacts. This has to be
+    ///         an instance of <c>KubernetesDiscoveryOptions</c>, <c>AkkaDiscoveryOptions</c>,
+    ///         <c>Ec2ServiceDiscoveryOptions</c>, <c>EcsServiceDiscoveryOptions</c>,
+    ///         or <c>ConfigServiceDiscoveryOptions</c>.
+    ///     </para>
+    ///     Note that if you're also using Akka.Discovery for ClusterBootstrap, in order for
+    ///     <see cref="ClusterClientDiscovery"/> to work, you <b>will need</b> to set
+    ///     <c>DiscoveryOptions.IsDefaultPlugin</c> to <c>false</c>
+    /// </summary>
+    public IHoconOption DiscoveryOptions { get; set; } = null!;
```
Aaronontheweb: Per our call, use `IDiscoveryOption` or whatever here

src/Akka.Cluster.Hosting/ClusterClientDiscoveryOptions.cs
```diff
@@ -0,0 +1,104 @@
...
+            sb.AppendLine($"resolve-timeout = {Timeout.ToHocon()}");
+        sb.AppendLine("}");
+
+        return sb.ToString();
+    }
+
+    private void Validate()
+    {
+        var type = DiscoveryOptions.GetType().Name;
+        if(!(type.Contains("ConfigServiceDiscoveryOptions")
```
Aaronontheweb: Rather than do stringly-typed stuff, use the `IDiscoveryOption` type we proposed on the call

Aaronontheweb (changes requested):
Suggestions

src/Akka.Cluster.Hosting/ClusterClientDiscoveryOptions.cs
Arkatufus: Done

src/Akka.Cluster.Hosting/ClusterClientDiscoveryOptions.cs
Arkatufus: Done

Aaronontheweb (approved):
LGTM

PR #480, opened 2024-07-26
Fix how ClusterShardingSettings were applied

### LINKED ISSUE #477: Multiple shardregion configuration using .WithShardRegion<T>(...) act similar
**Version Information**
Version of Akka.NET? 
Which Akka.NET Modules?

Akka.Cluster.Hosting, Version=1.5.25.0

**Describe the bug**

During setting up the ActorSystem using the .AddAkka(...) extension method and using the .WithShardRegion<T>(...) extension method to set up multiple shard regions, the ShardRegions act similar despite the different ShardOptions.
One of them is configured to passivate its entities the other configured to not, however both shard regions passivate their entities.

Using the ClusterSharding.Get(system).Start(...) way to set up shard regions works fine, they act different just as expected.

**To Reproduce**
Steps to reproduce the behavior: (see pictures below for more details)
1. Set up two simple ShardRegions using the .WithShardRegion<T>(...) extension method
2. Actors in ShardRegions should be ReceivePersistentActors
3. Each ShardRegions should have different Roles
4. One ShardRegion should have entity passivation ON, the other ShardRegion should have this feature turned OFF and have remember entities on
5. Each server should run the same code with both Roles
6. The test is made with 3 servers
7. When addressing the two ShardRegions, both freshly created actors will be pasivated, despite the different ShardOptions.

_Links to working reproductions on Github / Gitlab are very much appreciated_
https://github.com/Gaaaabor/Gaaaabor.Akka.Discovery.Docker/tree/feature/PossibleAkkaHostingBug

**Expected behavior**
The expected behavior would be one region with passivating its entities. the other is not.

**Actual behavior**
What actually happened and how did it differ from your expectations?
Both region passivated its entities, however only one of them should.

**Screenshots**
Setting up ShardRegions with .WithShardRegion<T>(...) , one without passivation the other with passivation:
image

Despite the configuration both shardregion's entities are passivated.
image

When the other method is used to set up the ShardRegions, one without passivation the other with passivation:
image

Both ShardRegion act as expected, one is not passivating the other is passivating its entities:
image

**Environment**
Are you running on Linux? Windows? Docker? Which version of .NET?
Running on Windows 10 using docker desktop and docker compose using .Net8.

**Additional context**

Arkatufus:
Fixes #477

Previously, settings were applied directly to the HOCON configuration, setting up a cluster shard region using `ShardOptions` would not work because only the last `ShardOptions` would be applied to the HOCON configuration.

## Changes

* Make `ShardOptions` to be applied separately per `WithShardRegion()` call.

> [!Note]
> Distributed data settings for all shard regions are still stored as a single HOCON setting (`akka.cluster.sharding.distributed-data`), they are not applied on a per region basis.

### DIFF
--- a/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
+++ b/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
@@ -279,7 +279,7 @@ namespace Akka.Cluster.Hosting
         /// <summary>
         ///     <para>
         ///         Settings for the Distributed Data replicator.
-        ///         The <see cref="DistributedData.Role"/> property is not used. The distributed-data
+        ///         The <see cref="ShardingDDataOptions.Role"/> property is not used. The distributed-data
         ///         role will be the same as <see cref="ShardOptions.Role"/>.
         ///         Note that there is one Replicator per role and it's not possible
         ///         to have different distributed-data settings for different sharding entity types.
@@ -287,6 +287,9 @@ namespace Akka.Cluster.Hosting
         ///     <b>NOTE</b> This setting is only used when <see cref="StateStoreMode"/> is set to
         ///     <see cref="Akka.Cluster.Sharding.StateStoreMode.DData"/>
         /// </summary>
+        [Obsolete("This property is not being applied to the ActorSystem anymore. " +
+                  "Use manual HOCON configuration to set \"akka.cluster.sharding.distributed-data\" values. " +
+                  "Since v1.5.27")]
         public ShardingDDataOptions DistributedData { get; } = new();
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -850,7 +874,7 @@ public static class AkkaClusterHostingExtensions
             IMessageExtractor messageExtractor,
             ShardOptions shardOptions)
         {
-            shardOptions.Apply(builder);
+            shardOptions.DistributedData.Apply(builder);
```
Aaronontheweb: Just don't worry about DData settings. Please remove this entirely from the API. Let users configure DData separately if needed. Only stick to what's in the `ClusterShardingSetting`s class for this.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -646,6 +641,35 @@ public static class AkkaClusterHostingExtensions
...
+        /// <param name="builder">
+        ///     The builder instance being configured.
+        /// </param>
+        /// <param name="shardOptions">
+        ///     The set of options for configuring <see cref="ClusterShardingSettings"/>
+        /// </param>
+        /// <returns>
+        ///     The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.
+        /// </returns>
+        public static AkkaConfigurationBuilder WithDefaultClusterShardingSettings(
```
Aaronontheweb: Delete this method - don't want it. Each `ShardRegion` should have its own configuration.

Aaronontheweb (changes requested):
Made some suggestions

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Arkatufus: how would the user set the `akka.cluster.sharding.distribute-data` settings then?

Aaronontheweb: If a situation arises where they would ever need to do it, they can do it through HOCON.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -860,7 +857,12 @@ public static class AkkaClusterHostingExtensions
             async Task Resolver(ActorSystem system, IActorRegistry registry, IDependencyResolver resolver)
             {
                 var props = entityPropsFactory(system, registry, resolver);
-                var settings = ClusterShardingSettings.Create(system);
+                var shardingConfig = shardOptions.ToConfig()
+                    .WithFallback(system.Settings.Config.GetConfig("akka.cluster.sharding"));
+                var coordinatorConfig = system.Settings.Config.GetConfig(
+                    shardingConfig.GetString("coordinator-singleton"));
+                
+                var settings = ClusterShardingSettings.Create(shardingConfig, coordinatorConfig);
```
Aaronontheweb, line 870: LGTM

Aaronontheweb (changes requested):
LGTM but add a regression spec to verify that two different ShardRegions in the same node can be configured with separate:

1. Remember entities
2. Passivization settings
3. Roles

You don't need to end2end this spec - just testing the `ClusterShardingSettings` output produced by consuming the `shardOptions.ToConfig()` should be fine.

Arkatufus:
Done

Aaronontheweb (approved):
LGTM

PR #489, opened 2024-08-29
Automatically add default Akka.Streams HOCON

Aaronontheweb:
## Changes

This is in response to a customer issue where the following happened:

1. Using `StreamRef`s
2. Transmits a `StreamRef` to a node that has never called any materialization functions
3. Serialization error due to missing serializer definitions.

In normal Akka.NET you'd have to fix this by manually appending `ActorMaterializer.DefaultConfig()` as a fallback. In Akka.Hosting we can just do this automatically at startup.

I added a reference to Akka.Streams in the default Akka.Hosting module. There's not a lot of value in making a custom Akka.Streams.Hosting package as "add this config" is the only method that would be in it. If there's a good reason why we shouldn't add it to the default Akka.Hosting package, please say so in a review.

## Checklist

### DIFF
--- a/src/Akka.Hosting/Akka.Hosting.csproj
+++ b/src/Akka.Hosting/Akka.Hosting.csproj
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <PackageReference Include="Akka.DependencyInjection" Version="$(AkkaVersion)" />
+    <PackageReference Include="Akka.Streams" Version="$(AkkaVersion)" />
     <PackageReference Include="Microsoft.Extensions.Hosting" Version="$(MicrosoftExtensionsVersion)" />
     <PackageReference Include="Microsoft.Extensions.Configuration" Version="$(MicrosoftExtensionsVersion)" />
   </ItemGroup>
--- a/src/Akka.Hosting/AkkaConfigurationBuilder.cs
+++ b/src/Akka.Hosting/AkkaConfigurationBuilder.cs
@@ -83,8 +83,8 @@ namespace Akka.Hosting
     {
         internal readonly string ActorSystemName;
         internal readonly IServiceCollection ServiceCollection;
-        internal readonly HashSet<SerializerRegistration> Serializers = new HashSet<SerializerRegistration>();
-        internal readonly List<Type> Extensions = new List<Type>();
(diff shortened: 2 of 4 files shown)


src/Akka.Hosting/AkkaConfigurationBuilder.cs
Aaronontheweb, line 1: Just some C# modernization

src/Akka.Hosting/Akka.Hosting.csproj
```diff
@@ -11,6 +11,7 @@
 
   <ItemGroup>
     <PackageReference Include="Akka.DependencyInjection" Version="$(AkkaVersion)" />
+    <PackageReference Include="Akka.Streams" Version="$(AkkaVersion)" />
```
Aaronontheweb, line 14: We need to move this project to centralized package management

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -58,6 +59,12 @@ public static IServiceCollection AddAkka(this IServiceCollection services, strin
         public static IServiceCollection AddAkka<T>(this IServiceCollection services, string actorSystemName, Action<AkkaConfigurationBuilder, IServiceProvider> builder) where T:AkkaHostedService
         {
             var b = new AkkaConfigurationBuilder(services, actorSystemName);
+            
+            // add the default Akka.Streams configuration by default - hurts nothing, but
+            // ensures that StreamRefs work correctly out of the box in case users
+            // haven't attempted to materialize a stream yet
+            b.AddHocon(ActorMaterializer.DefaultConfig(), HoconAddMode.Append);
```
Aaronontheweb, line 66: Added the HOCON here - it can be in Append mode because it's just the defaults - it can go at the end of the line.

src/Akka.Hosting/LoggerConfigBuilder.cs
Aaronontheweb, line 1: More C# modernization.

Aaronontheweb (commented):
Detailed my changes

Arkatufus (approved):
LGTM

PR #494, opened 2024-09-16
Make sure `AkkaHostedService` failures are always visible

### LINKED ISSUE #470: Errors that occur during launch of `AkkaHostedService` get eaten / not logged
**Version Information**
Version of Akka.NET? v1.5.15
Which Akka.NET Modules? Akka.Hosting

**Describe the bug**

Have the following code sample:

```csharp
public static AkkaConfigurationBuilder AddIssueTrackingActor(this AkkaConfigurationBuilder builder)
    {
        return builder
            .AddHocon(
                $"custom-issue-tracking-mailbox {{ mailbox-type: \"{typeof(IssueTrackerMailbox).AssemblyQualifiedName}\" }}",
                HoconAddMode.Append)
            .WithActors((system, registry) =>
            {
                Actor.Props props = Akka.Actor.Props.Create<IssueTrackingActor>()
                    .WithMailbox("custom-issue-tracking-mailbox");
                IActorRef actor = system.ActorOf(props, "issue-tracking-actor");

                //BUG: we never register the IssueTrackingActor with the ActorRegistry
            })
            .AddStartup((system, registry) =>
            {
                // BUG part 2: attempt to resolve non-existent entry from registry; exception thrown
                IActorRef actor = registry.Get<IssueTrackingActor>();
                
                // Send the issues in reverse order of importance.
                // Worth bearing in mind: totally possible that a lower-priority item might get processed
                // before the higher priority ones if the actor gets scheduled and runs immediately - it's
                // not going to sit there and wait for the higher priority items to arrive before starting.
                actor.Tell(new Issue("Issue 3", "Description 3", IssuePriority.Low));
                actor.Tell(new Issue("Issue 2", "Description 2", IssuePriority.Medium));
                actor.Tell(new Issue("Issue 1", "Description 1", IssuePriority.High));
                actor.Tell(new Issue("Issue 0", "Description 1", IssuePriority.High));
                
            });
    }
```

You can see where I detailed the bugs on there - the `registry.Get<IssueTrackingActor>();` call throws.

However, what I see when I run this without the debugger attached is just an immediate process exit with no indication that anything went wrong. This isn't a great user experience and we need to find some way of clearly communicating that a bad thing happened to the end-user.

Crashing the process and fast-exiting is probably ok, but the user has to be given an explanation for why this happened.

**Expected behavior**

Some kind of error message indicating what the problem was and why my process exited.

**Actual behavior**

Silent failure.

**Screenshots**

**Environment**

Running via Rider F5 without the debugger attached - this where I get the silent exit.

When running via `dotnet run` on the console I get this instead:

image

Seems like everything else startup up and ran ok there, but still no error message for the failed behavior.

Aaronontheweb:
## Changes

close #470 - application crashes and exits should always be explicitly logged someplace where they can be seen regardless of logging configuration.

## Checklist

* [x] Design discussion issue #470

### DIFF
--- a/src/Akka.Hosting/AkkaHostedService.cs
+++ b/src/Akka.Hosting/AkkaHostedService.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Runtime.ExceptionServices;
 using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
@@ -74,8 +75,13 @@ namespace Akka.Hosting
                 Logger.Log(LogLevel.Critical, ex, "Unable to start AkkaHostedService - shutting down application");
                 
                 // resolve https://github.com/akkadotnet/Akka.Hosting/issues/470 - never allow failures to be silent
-                Console.WriteLine($"Unable to start AkkaHostedService due to [{ex}] - shutting down application");
-                HostApplicationLifetime?.StopApplication();
+                Console.WriteLine($"Unable to start AkkaHostedService - shutting down application.\nCause: {ex}");
+                
+                // Best effort to perform a clean stop
+                var capturedException = ExceptionDispatchInfo.Capture(ex);
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(3));
+                await StopAsync(cts.Token);
+                capturedException.Throw();
             }
         }
 
--- a/src/Akka.Hosting/AkkaHostedService.cs
+++ b/src/Akka.Hosting/AkkaHostedService.cs
@@ -72,6 +72,9 @@ namespace Akka.Hosting
             catch (Exception ex)
             {
                 Logger.Log(LogLevel.Critical, ex, "Unable to start AkkaHostedService - shutting down application");
+                
+                // resolve https://github.com/akkadotnet/Akka.Hosting/issues/470 - never allow failures to be silent
+                Console.WriteLine($"Unable to start AkkaHostedService due to [{ex}] - shutting down application");
                 HostApplicationLifetime?.StopApplication();
             }
         }
--- a/src/Akka.Hosting.Tests/StartFailureSpec.cs
+++ b/src/Akka.Hosting.Tests/StartFailureSpec.cs
@@ -1,0 +1,52 @@
+﻿using System;
+using System.Threading.Tasks;
+using Akka.Actor;
+using FluentAssertions;
+using Microsoft.Extensions.DependencyInjection;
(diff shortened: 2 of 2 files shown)


src/Akka.Hosting/AkkaHostedService.cs
```diff
@@ -72,6 +72,9 @@ async Task TerminationHook()
             catch (Exception ex)
             {
                 Logger.Log(LogLevel.Critical, ex, "Unable to start AkkaHostedService - shutting down application");
+                
+                // resolve https://github.com/akkadotnet/Akka.Hosting/issues/470 - never allow failures to be silent
+                Console.WriteLine($"Unable to start AkkaHostedService due to [{ex}] - shutting down application");
```
Aaronontheweb: Ensure that this gets logged regardless of how the MSFT.EXT.Logger inputs have been configured

Arkatufus: I think for exception, it should always be added at the end, because the stack trace will screw up the string formatting.

Arkatufus: Or we can change that to "due to [{ex.Message}] - shutting down application.\nCause: {ex}"

Arkatufus: or maybe we should just re-throw the exception as a new exception and let MS.EXT.Hosting deal with it.

Arkatufus (changes requested):
I think this should be changed a bit

src/Akka.Hosting/AkkaHostedService.cs
```diff
@@ -74,8 +75,13 @@ async Task TerminationHook()
...
                 // resolve https://github.com/akkadotnet/Akka.Hosting/issues/470 - never allow failures to be silent
-                Console.WriteLine($"Unable to start AkkaHostedService due to [{ex}] - shutting down application");
-                HostApplicationLifetime?.StopApplication();
+                Console.WriteLine($"Unable to start AkkaHostedService - shutting down application.\nCause: {ex}");
+                
+                // Best effort to perform a clean stop
+                var capturedException = ExceptionDispatchInfo.Capture(ex);
+                using var cts = new CancellationTokenSource(TimeSpan.FromSeconds(3));
+                await StopAsync(cts.Token);
+                capturedException.Throw();
```
Aaronontheweb, line 84: LGTM

Aaronontheweb (commented):
Changes look good @Arkatufus

PR #511, opened 2024-11-06
Corrected name assigned to a singleton instance.

### LINKED ISSUE #509: WithSingleton extension method ignores singleton name specified in Akka HOCON
When we create singletons without using Akka.Hosting extensions, the actual singleton actor has path **//user/my_singleton_name/singleton**, i.e. if the name for a singleton manager actor is "my_singleton_name", then the singleton instance is created with "singleton" name, as specified in HOCON (akka.cluster.singleton.singleton-name). And this is correct behavior. However, when using Akka.Hosting WithSingleton extension, the singleton actor path becomes **//user/my_singleton_name/my_singleton_name**.

I believe the name gets overridden in this code (AkkaClusterHostingExtensions.cs):

```
                options ??= new ClusterSingletonOptions();
                var clusterSingletonManagerSettings =
                    ClusterSingletonManagerSettings.Create(system).WithSingletonName(singletonName);

```

First ClusterSingletonManagerSettings.Create(system) creates settings from akka.cluster.singleton config, but then the call to WithSingletonName(singletonName) overrides the value provided in akka.cluster.singleton.singleton-name, so the singleton path contains  the singleton basename (like my_singleton_name) twice instead of the base name with "/singleton" suffix.

object:
## Changes

This PR resolves #509. Assuming that _akka.cluster.singleton.singleton-name = singleton_ and the singleton is named **my_singleton**, an instance of a cluster singleton actor should have path **//user/my_singleton/singleton**, but extension methods **WithSingleton** and **WithSingletonProxy** assigned it to **//user/my_singleton/my_singleton**. This PR fixes this error.


Aaronontheweb:
As long as this doesn't break existing Akka.Hosting applications by changing the route of the singleton actor when it's being messaged via `ClusterSingletonProxy` instances I think this _should_ be ok. I'll check the test suite to see if it covers that.

Aaronontheweb:
Because the problem you can run into is that during an incremental upgrade to an existing Akka.Cluster application, this can temporarily break messaging.

Aaronontheweb:
Superseded via https://github.com/akkadotnet/Akka.Hosting/pull/520

PR #520, opened 2024-11-13
Fix `WithSingleton` and `WithSingletonProxy` API

### LINKED ISSUE #509: WithSingleton extension method ignores singleton name specified in Akka HOCON

Arkatufus:
Fixes #509

## Changes

* Allow for users to specify the singleton manager actor name

## Checklist

### DIFF
--- a/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
+++ b/src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
@@ -160,6 +160,53 @@ namespace Akka.Cluster.Hosting
         /// The interval between retries for acquiring the lease
         /// </summary>
         public TimeSpan? LeaseRetryInterval { get; set; }
+
+        /// <summary>
+        /// Interval at which the proxy will try to resolve the singleton instance.
+        /// </summary>
+        public TimeSpan? SingletonIdentificationInterval { get; set; }
+        
+        /// <summary>
+        /// Should the singleton proxy publish a warning if no singleton actor were found after a period of time
+        /// </summary>
+        public bool? LogSingletonIdentificationFailure { get; set; }
+        
+        /// <summary>
+        /// The period the proxy will wait until it logs a missing singleton warning, defaults to 1 minute
+        /// </summary>
+        public TimeSpan? SingletonIdentificationFailurePeriod { get; set; }
+        
+        internal ClusterSingletonManagerSettings ToManagerSettings(string singletonName, ActorSystem system)
+        {
+            var settings = ClusterSingletonManagerSettings.Create(system);
+            
+            var retry = LeaseRetryInterval ?? system.Settings.Config.GetTimeSpan("akka.cluster.singleton.lease-retry-interval");
+            var leaseSettings = LeaseImplementation is not null
+                ? new LeaseUsageSettings(LeaseImplementation.ConfigPath, retry)
+                : null;
+
+            return new ClusterSingletonManagerSettings(
+                singletonName: singletonName,
+                role: Role ?? settings.Role,
+                removalMargin: settings.RemovalMargin,
+                handOverRetryInterval: settings.HandOverRetryInterval,
+                leaseSettings: leaseSettings ?? settings.LeaseSettings,
+                considerAppVersion: false);
+        }
+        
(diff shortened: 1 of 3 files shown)


src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -160,6 +160,53 @@ public sealed class ClusterSingletonOptions
...
+        
+        /// <summary>
+        /// Should the singleton proxy publish a warning if no singleton actor were found after a period of time
+        /// </summary>
+        public bool? LogSingletonIdentificationFailure { get; set; }
+        
+        /// <summary>
+        /// The period the proxy will wait until it logs a missing singleton warning, defaults to 1 minute
+        /// </summary>
+        public TimeSpan? SingletonIdentificationFailurePeriod { get; set; }
```
Arkatufus, line 177: Added Akka 1.5.31 missing singleton detection feature

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -1222,49 +1279,89 @@ public static AkkaConfigurationBuilder WithSingleton<TKey>(
...
+        ///     <list type="bullet">
+        ///         <item>
+        ///             The complete singleton manager actor path name will be
+        ///             <c>akka://MyActorSystem/user/singletonManagerName/singletonName</c>
+        ///         </item>
+        ///         <item>
+        ///             The complete singleton proxy actor path name, if created, will be
+        ///             <c>akka://MyActorSystem/user/singletonName-proxy</c>
+        ///         </item>
+        ///     </list>
```
Arkatufus, line 1300: Added explanation on what the actor names will be in the XML-DOC

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -1222,49 +1279,89 @@ public static AkkaConfigurationBuilder WithSingleton<TKey>(
...
+        /// <returns>
+        ///     The same <see cref="AkkaConfigurationBuilder"/> instance originally passed in.
+        /// </returns>
+        public static AkkaConfigurationBuilder WithSingleton<TKey>(
+            this AkkaConfigurationBuilder builder,
+            string singletonManagerName,
+            string singletonName,
+            Func<ActorSystem, IActorRegistry, IDependencyResolver, Props> propsFactory,
+            ClusterSingletonOptions? options = null,
+            bool createProxyToo = true)
```
Arkatufus, line 1336: New API that allows users to specify the singleton manager actor name.

The resulting singleton path will be `akka://MyActorSystem/user/singletonManagerName/singletonName`

Old APIs are untouched, they will still generate the same singleton actor path `akka://MyActorSystem/user/singletonName/singletonName`

Arkatufus (commented):
Self-review

object (approved):
LGTM

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Aaronontheweb: :+1:

Aaronontheweb (approved):
lgtm

PR #527, opened 2024-11-19
Add global DData setting extension method

### LINKED ISSUE #512: No way to disable `DData.Durable` with Akka.Hosting
I can validate that this DOES disable DData from writing to disk:

```csharp
public static class DrawingSessionActorExtensions
{
    public static AkkaConfigurationBuilder AddDrawingSessionActor(this AkkaConfigurationBuilder builder,
        string clusterRoleName = ClusterConstants.DrawStateRoleName)
    {
        builder.WithShardRegion<DrawingSessionActor>("drawing-session",
                (system, registry, resolver) => s => resolver.Props<DrawingSessionActor>(s),
                new DrawingSessionActorMessageExtractor(), new ShardOptions()
                {
                    StateStoreMode = StateStoreMode.DData,
                    RememberEntities = true,
                    RememberEntitiesStore = RememberEntitiesStore.DData,
                    Role = clusterRoleName
                })
            // .WithDistributedData(options =>
            // {
            //     options.Durable = new DurableOptions() { Keys = [] }; // disable persistence
            // })
            .AddHocon("akka.cluster.sharding.distributed-data.durable.keys = []", HoconAddMode.Prepend);
        return builder;
    }
}
```

It looks like this is an Akka.Hosting issue with `options.Durable = new DurableOptions() { Keys = [] }; // disable persistence` not working as expected - and that's probably an easy fix.

_Originally posted by @Aaronontheweb in https://github.com/akkadotnet/akka.net/issues/7326#issuecomment-2460514548_

Arkatufus:
Fixes #512

## Changes

* [x] Make DData settings global instead of per-shard region.
* [x] Add unit test to make sure that all options are applied.

### DIFF
--- a/src/Akka.Cluster.Hosting.Tests/ShardOptionsSpecs.cs
+++ b/src/Akka.Cluster.Hosting.Tests/ShardOptionsSpecs.cs
@@ -1,0 +1,662 @@
+﻿using System;
+using System.Collections.Immutable;
+using System.Threading.Tasks;
+using Akka.Actor;
+using Akka.Cluster.Hosting.Tests.Lease;
+using Akka.Cluster.Sharding;
+using Akka.Cluster.Tools.Singleton;
+using Akka.Configuration;
+using Akka.Coordination;
+using Akka.DistributedData;
+using Akka.Hosting;
+using Akka.Persistence.Hosting;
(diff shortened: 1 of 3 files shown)


Aaronontheweb (approved):
LGTM

PR #602, opened 2025-05-28
[Persistence] Add supervisor strategy to journal and snapshot options

Arkatufus:
## Changes

Add `JournalOptions.SupervisorStrategy` and `SnapshotOptions.SupervisorStrategy` options that was added to Akka v1.5.42

### DIFF
--- a/src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApprovePersistence.verified.txt
+++ b/src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApprovePersistence.verified.txt
@@ -38,6 +38,7 @@ namespace Akka.Persistence.Hosting
         public bool IsDefaultPlugin { get; set; }
         public string PluginId { get; }
         public string? Serializer { get; set; }
+        public Akka.Actor.SupervisorStrategyConfigurator? SupervisorStrategy { get; set; }
         public abstract string Identifier { get; set; }
         protected virtual System.Text.StringBuilder Build(System.Text.StringBuilder sb) { }
         public Akka.Configuration.Config ToConfig() { }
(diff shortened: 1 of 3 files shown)


Aaronontheweb (commented):
Why not just assign a `SupervisionStrategy` directly?

Arkatufus:
Because the way it is set up in Akka, it requires a `SupervisorStrategyConfigurator`

Arkatufus:
https://github.com/akkadotnet/akka.net/blob/e2d1800fbf8c778157d4917f367edc2d784e5093/src/core/Akka.Persistence/Persistence.cs#L308-L309

https://github.com/akkadotnet/akka.net/blob/e2d1800fbf8c778157d4917f367edc2d784e5093/src/core/Akka/Actor/SupervisorStrategy.cs#L1002-L1021

Aaronontheweb (changes requested):
the goal is to assign a supervisor strategy
 - the `SupervisorConfigurator` or whatever it's called is a hack we have to have in order to get HOCON to work. 
 - We can just do the `SupervisorStrategy` directly here and generate the configurator for it. That would be the DX-friendly way to do this.

Arkatufus:
Yes, but that is not how the supervisor strategy being loaded from HOCON the HOCON loader __requires__ that the type being passed in as a `SupervisorStrategyConfigurator` or it __will__ throw. If you want to do that, we have to change how it is done in Akka.Persistence first.

Arkatufus:
We can't just "generate" a configurator because it still needs to be loaded from HOCON in the end. And as you said, "the SupervisorConfigurator or whatever it's called is a hack we have to have in order to get HOCON to work". You mean actually use code generator to make a class during compilation time?

PR #632, opened 2025-07-23
Fix IndexOutOfRangeException in Configuration to HOCON adaptor

### LINKED ISSUE #631: HOCON adapter: `ExpandKeys` errors when key length == 0
**Version Information**
Version of Akka.NET? Akka.Hosting
Which Akka.NET Modules? at least v1.5.44, if not earlier

**Describe the bug**

https://github.com/akkadotnet/Akka.Hosting/blob/bbabe2a7e12c00a8a23a8ae02fe0d3c7aa217d74/src/Akka.Hosting/Configuration/ConfigurationHoconAdapter.cs#L40-L70

If `keys.length == 0`, that scenario is unhandled inside this method and we try to access `keys[0]`, which results in an `IndexOutOfRange` exception.

This behavior appears to be dependent on the types of environment variables that are present - in my case, it's these ones:

Image

**Expected behavior**

Parser should not die on `IConfigurationSection` instances that have no key / path.

**Actual behavior**

Throws an `IndexOutOfRange` exception currently.

Arkatufus:
Fixes #631

## Changes

Detect cases where MS.EXT.Configuration.EnvironmentVariables mistook double underscore as a section separator when it should have been part of the variable name


Aaronontheweb (approved):
LGTM

PR #633, opened 2025-07-23
Microsoft.Extensions.Diagnostics.HealthChecks integration and automation

Aaronontheweb:
## Changes

Successor to https://github.com/akkadotnet/Akka.Hosting/pull/629.

Completely replaces [Akka.HealthChecks](https://github.com/petabridge/akkadotnet-healthcheck)

## Checklist

* [x] I have added website documentation for this feature.

### Problems with Akka.HealthChecks

There are a few major problems with Akka.HealthChecks:

1. **Hair triggers that can inadvertently nuke an otherwise functioning Akka.NET cluster**: see https://github.com/petabridge/akkadotnet-healthcheck/issues/278 and https://github.com/petabridge/akkadotnet-healthcheck/issues/237 - we have had to do constant firefighting to get these right over the years. The fundamental problem is that Akka.HealthChecks tries to do too much and doesn't give the underlying systems time to recover, resulting in wild swings in availability. The new design approaches these things much more carefully and, generally, tries to use fewer, more meaningful health checks.
2. **Clunky and awkward to configure** - you have to configure health checks in three places, you have to install several different NuGet packages, and there are a lot of settings + moving parts involved. A lot of this is legacy baggage from before Microsoft.Extensions.Diagnostics.HealthChecks existed.
3. **Difficult to customize** - writing custom health checks with Akka.HealthCheck is... arduous, to say the least.

### How This PR Solves Them

1. **No additional packages** - Akka.Hosting now takes a direct dependency on Microsoft.Extensions.Diagnostics.HealthChecks and exposes APIs for configuring Akka.NET-specific health checks that will be registered with the `HealthCheckService`.
2. **Automatic registration of health checks with `HealthCheckService`** - if you call any of the `WithHealthCheck` overloads on the `AkkaConfigurationBuilder`, those types will automatically be registered with the `Microsoft.Extensions.Diagnostics.HealthChecks.HealthCheckService` and they will "just work." No additional API calls required - happens automatically.
3. **Automatic health checks for core Akka and Akka.Cluster** - we ship with two built in checks out of the box: an `ActorSystem` liveness check - if it's dead, the check fails. And an Akka.Cluster "have we joined the cluster yet?" readiness check. You can see what the outputs of this look like by running:

```shell
dotnet run --project src/Examples/Akka.Hosting.Asp.LoggingDemo/Akka.Hosting.Asp.LoggingDemo.csproj
```

This will expose the output of both of these healthchecks as pretty-printed JSON at http://localhost:5000/healthz

```json
{
  "status": "Healthy",
  "totalDuration": "00:00:00.0002317",
  "checks": [
    {
      "name": "ActorSystem Available",
      "status": "Healthy",
      "duration": "00:00:00.0000138",
      "description": "ActorSystem is running.",
      "tags": [
        "akka"
      ],
      "data": {}
    },
    {
      "name": "cluster.join",
      "status": "Healthy",
      "duration": "00:00:00.0000248",
      "description": "Successfully joined Akka.NET cluster after [0:00:06.3427821].",
      "tags": [
        "ready",
        "akka.cluster",
        "akka"
      ],
      "data": {}
    }
  ]
}
```

4. **Easy to customize** - we expose several types of helper methods, all called `WithHealthCheck`, that make it very easy to define a health check. For example:

```csharp
builder.WithHealthCheck("FooActor alive", async (system, registry, cancellationToken) =>
        {
            /*
             * N.B. CancellationToken is set by the call to MSFT.EXT.DIAGNOSTICS.HEALTHCHECK,
             * so that value could be "infinite" by default.
             *
             * Therefore, it might be a really, really good idea to guard this with a non-infinite
             * timeout via a LinkedCancellationToken here.
             */
            try
            {
                var fooActor = await registry.GetAsync<FooActor>(cancellationToken);

                try
                {
                    var r = await fooActor.Ask<ActorIdentity>(new Identify("foo"), cancellationToken: cancellationToken);
                    if (r.Subject.IsNobody())
                        return HealthCheckResult.Unhealthy("FooActor was alive but is now dead");
                }
                catch (Exception e)
                {
                    return HealthCheckResult.Degraded("FooActor found but non-responsive", e);
                }
            }
            catch (Exception e2)
            {
                return HealthCheckResult.Unhealthy("FooActor not found in registry", e2);
            }

            return HealthCheckResult.Healthy("fooActor found and responsive");
        });
```

A basic lambda function can handle it - the only reason this function is as complicated as this is because I wanted to include _different reasons for why it failed_ in the description. You can also implement your own `IAkkaHealthCheck` and pass it into a `WithHealthCheck` method that way too.


src/Akka.Cluster.Hosting/AkkaClusterHealthCheck.cs
```diff
@@ -0,0 +1,203 @@
...
+/// A programmable health check for Akka.Cluster. Make sure you read the documentation carefully.
+///
+/// You often will not need this - Akka.NET's Split Brain Resolvers (see https://getakka.net/articles/clustering/split-brain-resolver.html)
+/// typically handle most scenarios that require intervention automatically; they're enabled by default; and they will automatically
+/// kill unhealthy Akka.NET processes in order to ensure that there is a single, unified Akka.NET cluster afterwards.
+///
+/// This component is useful for scenarios where you want to trigger node-death when your cluster is missing enough
+/// nodes to fulfill its duties and can't really function anyway.
+/// </summary>
+internal sealed class AkkaClusterHealthCheck : IAkkaHealthCheck
```
Aaronontheweb, line 33: Not ready to expose this publicly yet as it needs testing / conceptual validation first, but this was my take on safely exposing a structured way to evaluate the health of the cluster without running the risk of completely nuking it by accident.

src/Akka.Cluster.Hosting/AkkaClusterHealthCheck.cs
```diff
@@ -0,0 +1,203 @@
...
+
+        // don't know enough
+        return Task.FromResult(HealthCheckResult.Healthy(data: cluster.State.DumpClusterState()));
+    }
+}
+
+/// <summary>
+/// Checks to see if we've joined a cluster.
+/// </summary>
+internal sealed class AkkaClusterReadinessCheck : IAkkaHealthCheck
```
Aaronontheweb: Akka.Cluster readiness check implementation - gets automatically registered and executed at launch. Returns `Unhealthy` until we successfully join.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -636,6 +636,10 @@ internal static AkkaConfigurationBuilder BuildClusterHocon(
             // prepend the composed configuration
             builder.AddHocon(sb.ToString(), HoconAddMode.Prepend);
 
+            // add the default cluster readiness check
+            builder.WithHealthCheck(new AkkaHealthCheckRegistration("cluster.join", new AkkaClusterReadinessCheck(),
```
Aaronontheweb: Registers the automatic cluster health check - applies a `ready` tag to it so it can be included in a `healthz/ready` filter like Microsoft recommends in its documentation: https://learn.microsoft.com/en-us/aspnet/core/host-and-deploy/health-checks?view=aspnetcore-9.0#separate-readiness-and-liveness-probes

src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -122,6 +125,7 @@ public AkkaConfigurationBuilder(IServiceCollection serviceCollection, string act
         {
             ServiceCollection = serviceCollection;
             ActorSystemName = actorSystemName;
+            AddBuiltInHealthChecks();
```
Aaronontheweb: I don't anticipate adding more built-in health checks to core Akka.NET, but this method is where they would go.

src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -332,6 +336,16 @@ public AkkaConfigurationBuilder WithExtension<T>() where T : IExtensionId
 
             return this;
         }
+
+        /// <summary>
+        /// Registers an <see cref="AkkaHealthCheckRegistration"/> with the <see cref="AkkaConfigurationBuilder"/>.
+        /// </summary>
+        /// <param name="registration">The health check registration.</param>
+        public AkkaConfigurationBuilder WithHealthCheck(AkkaHealthCheckRegistration registration)
```
Aaronontheweb, line 345: Core registration method for Akka.NET health checks - the other two flavors of this are extension methods that make it easier to produce `AkkaHealthCheckRegistration` instances.

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -256,5 +260,38 @@ public static AkkaConfigurationBuilder WithActors(this AkkaConfigurationBuilder
...
+        /// Registers a named <see cref="IAkkaHealthCheck"/> instance.
+        /// </summary>
+        /// <param name="builder">The builder instance being configured.</param>
+        /// <param name="name">The unique name for this health check.</param>
+        /// <param name="healthCheck">The health check implementation.</param>
+        /// <remarks>
+        /// If you need more detailed configuration for a health check, such as tags or default failure status,
+        /// please use the <see cref="AkkaConfigurationBuilder.WithHealthCheck(AkkaHealthCheckRegistration)"/> method.
+        /// </remarks>
+        public static AkkaConfigurationBuilder WithHealthCheck(this AkkaConfigurationBuilder builder, string name, IAkkaHealthCheck healthCheck)
```
Aaronontheweb, line 274: Quick method for adding an `IHealthCheck` + its name

src/Akka.Hosting/AkkaHostingExtensions.cs
```diff
@@ -256,5 +260,38 @@ public static AkkaConfigurationBuilder WithActors(this AkkaConfigurationBuilder
...
+        /// Registers a named <see cref="IAkkaHealthCheck"/> instance.
+        /// </summary>
+        /// <param name="builder">The builder instance being configured.</param>
+        /// <param name="name">The unique name for this health check.</param>
+        /// <param name="healthCheck">A health checking function.</param>
+        /// <remarks>
+        /// If you need more detailed configuration for a health check, such as tags or default failure status,
+        /// please use the <see cref="AkkaConfigurationBuilder.WithHealthCheck(AkkaHealthCheckRegistration)"/> method.
+        /// </remarks>
+        public static AkkaConfigurationBuilder WithHealthCheck(this AkkaConfigurationBuilder builder, string name,
```
Aaronontheweb, line 290: Allows users to build delegate-based health checks, so they can avoid the ceremony of having to design an `IAkkaHealthCheck` class if necessary.

src/Akka.Hosting/HealthChecks/ActorSystemCheck.cs
```diff
@@ -0,0 +1,23 @@
...
+using System.Threading.Tasks;
+using Akka.Actor;
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Akka.Hosting.HealthChecks;
+
+/// <summary>
+/// Checks to see if the <see cref="ActorSystem"/> is alive or not.
+/// </summary>
+internal sealed class ActorSystemLivenessCheck : IAkkaHealthCheck
```
Aaronontheweb: `ActorSystem` liveness check - returns `Unhealthy` if the `ActorSystem` is dead.

src/Akka.Hosting/HealthChecks/AkkaHealthCheckExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+
+internal static class AkkaHealthCheckExtensions
+{
+    public const string AkkaTag = "akka";
+    
+    /// <summary>
+    /// Converts an <see cref="AkkaHealthCheckRegistration"/> to a <see cref="HealthCheckRegistration"/>
+    /// </summary>
+    /// <param name="registration">the original Akka.NET health check registration.</param>
+    public static HealthCheckRegistration ToHealthCheckRegistration(this AkkaHealthCheckRegistration registration)
```
Aaronontheweb, line 16: Converts the `AkkaHealthCheckRegistration` into a regular MSFT/EXT/DIAG/HealthChecks `HealthCheckRegistration`. We let the MSFT services do all of the heavy lifting - we just wrote this so we could:

1. Define AKka.NET health checks succinctly, using our own parameters and types and
2. Adapt the context to include some Akka-specific concepts, such as reference to the `ActorSystem`.

src/Akka.Hosting/HealthChecks/HealthCheckAdapter.cs
```diff
@@ -0,0 +1,32 @@
...
+using Microsoft.Extensions.Diagnostics.HealthChecks;
+
+namespace Akka.Hosting.HealthChecks;
+
+/// <summary>
+/// INTERNAL API
+///
+/// Adapter wrapper around the <see cref="IAkkaHealthCheck"/> to make it API-compatible with <see cref="IHealthCheck"/>
+/// </summary>
+internal sealed class HealthCheckAdapter : IHealthCheck
```
Aaronontheweb, line 13: Wraps around an `IAkkaHealthCheck` and surfaces it as an `IHealthCheck`

src/Akka.Hosting/IAkkaHealthCheck.cs
```diff
@@ -0,0 +1,163 @@
...
+namespace Akka.Hosting;
+
+/// <summary>
+/// The registration record used to track this Akka.HealthCheck inside the <see cref="AkkaConfigurationBuilder"/>
+/// </summary>
+/// <remarks>
+/// These will all get converted into <see cref="HealthCheckRegistration"/>s by the <see cref="AkkaConfigurationBuilder"/>,
+/// and a default "akka.net" tag will be added to each of those registrations for filtering purposes.
+/// </remarks>
+public sealed class AkkaHealthCheckRegistration
```
Aaronontheweb, line 17: Very similar to a `HealthCheckRegistration` but without any direct DI dependencies

src/Akka.Hosting/IAkkaHealthCheck.cs
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+}
+
+/// <summary>
+/// Akka.NET health check invocation context. Provides access to the current Akka environment
+/// along with the settings for this health check, via <see cref="Registration"/>.
+/// </summary>
+public sealed class AkkaHealthCheckContext
```
Aaronontheweb, line 124: This is the more important differentiator - the context object. We need to make sure that the live running `ActorSystem` and its `ActorRegistry` get exposed to the health checks directly.

src/Akka.Hosting/IAkkaHealthCheck.cs
```diff
@@ -0,0 +1,163 @@
...
+    /// The health check registration data.
+    /// </summary>
+    /// Expected to be not-null in live environments, but can be null during testing.
+    public HealthCheckRegistration Registration { get; set; } = default!;
+}
+
+/// <summary>
+/// Healthcheck aimed at testing the health of Akka.NET-specific resources.
+/// </summary>
+public interface IAkkaHealthCheck
```
Aaronontheweb, line 151: The health check interface definition - very conceptually similar to `IHealthCheck`, the only difference is the context type.

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
This is ready for review - I'll contribute documentation once there's buy-in on the design.

src/Akka.Hosting.Tests/HealthChecks/HealthChecksSpec.cs
```diff
@@ -0,0 +1,125 @@
...
+
+                try
+                {
+                    var r = await fooActor.Ask<ActorIdentity>(new Identify("foo"), cancellationToken: cancellationToken);
+                    if (r.Subject.IsNobody())
+                        return HealthCheckResult.Unhealthy("FooActor was alive but is now dead");
+                }
+                catch (Exception e)
+                {
+                    return HealthCheckResult.Degraded("FooActor found but non-responsive", e);
```
alexaka1, line 51: Not too familiar with Akka terminology, but to me degraded means that things are working, but maybe the latency is outside of margin. Or core functionality is working, but some services may not. Ie.: Steam multiplayer is working, but the Chat widget not working would be arguably a degraded experience, but not unhealthy.

Is the actor being non-responsive not an unhealthy situation? Are things still working?

IIRC healthchecks are configurable at registration, so one could set `I want degraded to mean unhealty in Akka` so it's not that big of a deal, but I'd have to check.

Aaronontheweb: In this case it might mean degraded because the actor was too busy to respond on-time (large backlog of messages, CPU was busy, blocked on a long-running operation, etc) - so "downgraded" just reflects the uncertainty of that situation.

But this is just an example - if you use a full-blown `AkkaHealthCheckRegistration` can you have more control over things like the default failure status code you might want to return.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
```diff
@@ -663,6 +663,10 @@ public static AkkaConfigurationBuilder WithClustering(
             ClusterOptions? options = null)
         {
             var hoconBuilder = BuildClusterHocon(builder, options);
+            
+            // add the default cluster readiness check
+            builder.WithHealthCheck(new AkkaHealthCheckRegistration("akka.cluster.join", new AkkaClusterReadinessCheck(),
+                HealthStatus.Unhealthy, ["ready", "akka.cluster"]));
```
Arkatufus: Not sure if we want to do this. If we're still unsure that this works as intended and it still needs testing, why are we turning this on by default and exposing it to the user? If something were to go wrong, there's no way for the user to change this behavior since `AkkaClusterReadinessCheck` is internal selaed.

src/Akka.Cluster.Hosting/AkkaClusterHealthCheck.cs
Arkatufus: This returns healthy even after the node was removed from the cluster?

Arkatufus (changes requested):
Left a few questions

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Aaronontheweb: it definitely works - you can test it via the local demo.

src/Akka.Cluster.Hosting/AkkaClusterHealthCheck.cs
Aaronontheweb: By design - it's only supposed to be a readiness check to signal to the load-balancer if this node is ready to accept traffic. Its job is done afterwards.

src/Akka.Cluster.Hosting/AkkaClusterHostingExtensions.cs
Arkatufus: What I meant is that, if this has some weird side effect, say in microsoft fabric, and the user needs to modify and/or disable it, they can't, since it is an `internal sealed` class and we're forcing it on right here.

src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApproveCluster.verified.txt
```diff
@@ -41,6 +42,16 @@ namespace Akka.Cluster.Hosting
         public static Akka.Hosting.AkkaConfigurationBuilder WithSingletonProxy<TKey>(this Akka.Hosting.AkkaConfigurationBuilder builder, string singletonManagerName, string singletonName, Akka.Cluster.Hosting.ClusterSingletonOptions? options = null) { }
         public static Akka.Hosting.AkkaConfigurationBuilder WithSingletonProxy<TKey>(this Akka.Hosting.AkkaConfigurationBuilder builder, string singletonName, Akka.Cluster.Hosting.ClusterSingletonOptions? options = null, string? singletonManagerPath = null) { }
     }
+    public sealed class AkkaClusterReadinessCheck : Akka.Hosting.IAkkaHealthCheck
```
Aaronontheweb, line 45: Made the built-in health checks public

src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApproveCluster.verified.txt
```diff
@@ -3,6 +3,7 @@ namespace Akka.Cluster.Hosting
 {
     public static class AkkaClusterHostingExtensions
     {
+        public static Akka.Hosting.AkkaConfigurationBuilder WithAkkaClusterReadinessCheck(this Akka.Hosting.AkkaConfigurationBuilder builder) { }
```
Aaronontheweb, line 6: Added an explicit method for enabling the readiness check

src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApproveCluster.verified.txt
```diff
@@ -65,6 +76,7 @@ namespace Akka.Cluster.Hosting
     {
         public ClusterOptions() { }
         public string? AppVersion { get; set; }
+        public bool ClusterReadyCheck { get; set; }
```
Aaronontheweb, line 79: Added a config-based option for automatically enabling the `WithAkkaClusterReadinessCheck`

src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApproveCore.verified.txt
```diff
@@ -214,6 +239,14 @@ namespace Akka.Hosting.Coordination
         public abstract void Apply(Akka.Hosting.AkkaConfigurationBuilder builder, Akka.Actor.Setup.Setup? setup = null);
     }
 }
+namespace Akka.Hosting.HealthChecks
+{
+    public sealed class ActorSystemLivenessCheck : Akka.Hosting.IAkkaHealthCheck
```
Aaronontheweb, line 244: Made the `ActorSystemLivenessCheck` public

src/Akka.Hosting.API.Tests/verify/CoreApiSpec.ApproveCore.verified.txt
```diff
@@ -71,10 +89,13 @@ namespace Akka.Hosting
         public static Akka.Hosting.AkkaConfigurationBuilder AddSetup(this Akka.Hosting.AkkaConfigurationBuilder builder, Akka.Actor.Setup.Setup setup) { }
         public static Akka.Hosting.AkkaConfigurationBuilder WithActorAskTimeout(this Akka.Hosting.AkkaConfigurationBuilder builder, System.TimeSpan timeout) { }
         public static Akka.Hosting.AkkaConfigurationBuilder WithActorRefProvider(this Akka.Hosting.AkkaConfigurationBuilder builder, Akka.Actor.ProviderSelection providerSelection) { }
+        public static Akka.Hosting.AkkaConfigurationBuilder WithActorSystemLivenessCheck(this Akka.Hosting.AkkaConfigurationBuilder builder) { }
```
Aaronontheweb, line 92: Added a method to explicitly register the `ActorSystemLivenessCheck` as it's not enabled by default any longer

src/Akka.Hosting.Tests/HealthChecks/HealthChecksSpec.cs
alexaka1: Ah I missed this was a test. I was on the mobile webview and with about 70 pixel width for file names I didn't see that. My bad.

src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -84,6 +86,9 @@ public sealed class AkkaConfigurationBuilder
         internal readonly string ActorSystemName;
         internal readonly IServiceCollection ServiceCollection;
         internal readonly HashSet<SerializerRegistration> Serializers = new();
+        
+        // we use a name / registration dictionary to make health check registrations unique by name
+        internal readonly Dictionary<string, AkkaHealthCheckRegistration> HealthChecks = new();
```
Aaronontheweb, line 91: Reorganized the internal storage of the health check system to use a dictionary and make the name of the health check the primary key - this is to help ensure that there can't be duplicate entries of the same health checks using our built-in methods or any user-defined ones.

Aaronontheweb: Since equality is done by reference on the `AkkaHealthCheckRegistration` instances

Aaronontheweb (commented):
Removed the automatic registration of any built-in health checks, made the built-in health check implementations `public`, and added methods for explicitly opting-in to turning those health checks on.

Arkatufus (approved):
LGTM

src/Akka.Hosting.Tests/HealthChecks/HealthChecksSpec.cs
Aaronontheweb: @alexaka1 no problem - I appreciate you taking the time to review this code!

Aaronontheweb:
Going to add documentation to this PR and then merge it into a newly created `feature/healthcheck` branch so we can deploy a beta of it.

PR #646, opened 2025-08-19
ensure TestActor is created on main test thread context

### LINKED ISSUE #458: TestActor behaves unpredictably - looks like a static variable coupled with thread-race issue
This issue has a bit of a history in discord: https://discord.com/channels/974500337396375553/974500337954222133/1245403250841096325

Here is a copy of the discussion: 

brah.mcdude — Yesterday at 6:47 PM
hi all. 
i am running tests using Akka.Hosting.TestKit. I have a test that mostly passes. But randomly fails. I don't know why. 
Here is the code: 

```c#
public class Spec(ITestOutputHelper output) : TestKit(output, LogLevel.Debug)
{
    private IActorRef? myActor;

    [Theory]
    [InlineData(1)]
    [InlineData(2)]
    [InlineData(3)]
    [InlineData(4)]
    [InlineData(5)]
    [InlineData(6)]
    [InlineData(7)]
    public async Task Spec1(int _)
    {
        myActor.Tell(new { });
        await ExpectMsgAsync<object>();
    }

    {
        builder.WithActors((system, registry, resolver) =>
        {
            var props = resolver.Props<MyActor>();
            myActor = system.ActorOf(props, nameof(MyActor));
            registry.Register<MyActor>(myActor);
        });
    }

    public class MyActor : UntypedActor
    {
        protected override void OnReceive(object message)
        {
            Sender.Tell(new { });
        }
    }
}

```

Here is the log of the failure: 
        Failed: Timeout 00:00:03 while waiting for a message of type System.Object
brah.mcdude — Yesterday at 7:51 PM
ok. found out the cause of the issue. for some reason, we need a probe: 
```
        var probe = CreateTestProbe();
        myActor.Tell(new { }, probe);
        await probe.ExpectMsgAsync<object>();
```

Should i report it as a bug?

Aaronontheweb — Yesterday at 7:53 PM
I think I know what your issue is
https://github.com/akkadotnet/Akka.Hosting/issues/237

GitHub
[Akka.Hosting.TestKit: `TestActor` is no longer the implicit sender ...](https://github.com/akkadotnet/Akka.Hosting/issues/237)
Version Information Version of Akka.NET? 1.0.3 Which Akka.NET Modules? Akka.Hosting.TestKit Describe the bug foreach (var i in Enumerable.Range(0, 2)) { var update1 = Dsl.Update(CounterKey, GCounte...
Akka.Hosting.TestKit: `TestActor` is no longer the implicit sender ...

a very annoying bug with the TestKit
in Akka.Hosting

brah.mcdude — Today at 1:08 AM
unlikely. my tests mostly pass. 80% of the time. it can't be something like "wrong sender". as that would always fail.

brah.mcdude — Today at 1:58 PM
my intuition is that there are static variables. maybe the test-actor is static. and that only multiple tests running asynchronously are causing the issue.

brah.mcdude — Today at 3:18 PM
i'll open a bug to discuss this.

Aaronontheweb — Today at 4:40 PM
that actor context bug is a thread static variable
and the conditions under which you can lose that context can be racy
so I suspect that's what it is - have run into it myself using the Akka.Hosting.TestKit before

Aaronontheweb:
## Changes

fix #458 - this should ensure that the `TestActor` is present on the main implicit thread context the same way that it is in the normal `TestKit`s.

## Checklist

* [x] Design discussion issue #458

### DIFF
--- a/src/Akka.Hosting.TestKit/TestKit.cs
+++ b/src/Akka.Hosting.TestKit/TestKit.cs
@@ -115,7 +115,8 @@ namespace Akka.Hosting.TestKit
         {
             var extSystem = (ExtendedActorSystem)system;
             var logger = extSystem.SystemActorOf(Props.Create(() => new TestKitLoggerFactoryLogger()), "log-test");
-            await logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream));
+            // Add timeout to prevent deadlock when multiple tests run in parallel
+            await logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TimeSpan.FromSeconds(5));
         }
 
         protected virtual Config? Config { get; } = null;
@@ -161,7 +162,35 @@ namespace Akka.Hosting.TestKit
             
             var system = _host.Services.GetRequiredService<ActorSystem>();
             var registry = _host.Services.GetRequiredService<ActorRegistry>();
-            base.InitializeTest(system, (ActorSystemSetup)null!, null, null);
+            
+            // Initialize TestActor on the current synchronization context to preserve
+            // the implicit sender thread context (fixes #458)
+            // We need to capture the current synchronization context before any await
+            var currentContext = SynchronizationContext.Current;
+            if (currentContext != null)
+            {
+                // Post the initialization to the current context to ensure it runs on the correct thread
+                var tcs = new TaskCompletionSource<bool>();
+                currentContext.Post(_ =>
+                {
+                    try
+                    {
+                        base.InitializeTest(system, (ActorSystemSetup)null!, null, null);
+                        tcs.SetResult(true);
+                    }
+                    catch (Exception ex)
+                    {
+                        tcs.SetException(ex);
+                    }
+                }, null);
+                await tcs.Task;
+            }
+            else
+            {
+                // No synchronization context, run directly
+                base.InitializeTest(system, (ActorSystemSetup)null!, null, null);
+            }
+            
             registry.Register<TestProbe>(TestActor);
             
             if (this is not INoImplicitSender && InternalCurrentActorCellKeeper.Current is null)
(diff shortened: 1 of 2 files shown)


src/Akka.Hosting.TestKit/TestKit.cs
```diff
@@ -102,19 +102,6 @@ private void InternalConfigureServices(HostBuilderContext context, IServiceColle
                     });
                 }
 
-                builder.StartActors((system, registry) =>
-                {
-                    try
-                    {
-                        base.InitializeTest(system, (ActorSystemSetup)null!, null, null);
```
Aaronontheweb, line 109: I think this was part of our problem - the synchronization context we use to keep track of the implicit `TestActor` was being created, possibly but not always, on a separate thread from the main test execution context.

src/Akka.Hosting.TestKit/TestKit.cs
```diff
@@ -172,6 +159,11 @@ public async Task InitializeAsync()
 
             await _initialized.Task;
             
+            var system = _host.Services.GetRequiredService<ActorSystem>();
+            var registry = _host.Services.GetRequiredService<ActorRegistry>();
+            base.InitializeTest(system, (ActorSystemSetup)null!, null, null);
```
Aaronontheweb: By moving the call here we should be back in the normal `SynchronizationContext` the test runner uses, which should force the `TestActor` to always be the expected implicit sender, rather than randomly failing sometimes.

Aaronontheweb (commented):
Detailed my changes

Aaronontheweb:
I've pushed a fix for the race condition that was causing the test failures. 

## The Issue
The race condition occurred when multiple tests run in parallel. The logger initialization was using `Ask` without a timeout, causing deadlocks when multiple tests tried to initialize simultaneously.

## The Fix
1. Added a 5-second timeout to the `LoggerHook` Ask operation
2. Implemented proper SynchronizationContext preservation for TestActor initialization

The `TestActorStartupDeadlockSpec.parallel_host_start_should_not_deadlock` test now passes in ~7 seconds instead of timing out after 20 seconds.

Arkatufus (approved):
LGTM

PR #651, opened 2025-08-25
Fix race condition in TestKit when actors depend on IRequiredActor<TestProbe>

Aaronontheweb:
## Summary
Fixes a race condition in TestKit that caused deadlocks when actors depend on `IRequiredActor<TestProbe>` and multiple TestKit instances run in parallel.

## Problem
After merging PR #650, CI/CD tests began failing intermittently. Investigation revealed that when actors created via `WithActors` tried to resolve `IRequiredActor<TestProbe>` in their `PreStart` method, they could deadlock if TestProbe wasn't registered yet. This became apparent when running 30 parallel TestKit instances.

## Root Cause
The execution order was:
1. ActorSystem created
2. User's `WithActors` runs (creates actors immediately) 
3. TestProbe registered in `AddStartup` (too late\!)

When actors' `PreStart` ran before step 3, they would block forever waiting for TestProbe.

## Solution
- Move TestProbe registration from `AddStartup` to `StartActors` 
- Ensure it's added BEFORE user's `ConfigureAkka` method
- This guarantees TestProbe is registered before any user-defined actors start

## Additional Fixes
- Made `TestKitLoggerFactoryLogger` initialization fire-and-forget to avoid blocking
- Ensured implicit sender context is properly set on test threads

## Test Results
- ✅ All 279 TestKit tests pass
- ✅ DiPropsFailTest passes (original implicit sender issue)
- ✅ TestActorStartupDeadlockSpec passes with 30 concurrent hosts
- ✅ No regressions in existing functionality

Fixes issues introduced after #650 was merged.

### DIFF
--- a/src/Akka.Hosting.TestKit/TestKit.cs
+++ b/src/Akka.Hosting.TestKit/TestKit.cs
@@ -126,12 +126,15 @@ namespace Akka.Hosting.TestKit
             });
         }
 
-        internal virtual async Task LoggerHook(ActorSystem system, IActorRegistry registry)
+        internal virtual Task LoggerHook(ActorSystem system, IActorRegistry registry)
         {
             var extSystem = (ExtendedActorSystem)system;
-            var logger = extSystem.SystemActorOf(Props.Create(() => new TestKitLoggerFactoryLogger()), "log-test");
-            // Add timeout to prevent deadlock when multiple tests run in parallel
-            await logger.Ask<LoggerInitialized>(new InitializeLogger(system.EventStream), TimeSpan.FromSeconds(5));
+            var loggerName = $"log-test-{Guid.NewGuid():N}";
+            var logger = extSystem.SystemActorOf(Props.Create(() => new TestKitLoggerFactoryLogger()), loggerName);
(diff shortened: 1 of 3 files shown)


src/Akka.Hosting.TestKit/TestKit.cs
```diff
@@ -116,7 +116,24 @@ private void InternalConfigureServices(HostBuilderContext context, IServiceColle
                         await LoggerHook(system, registry);
                     });
                 }
+                
+                // Register TestProbe using StartActors (not AddStartup) so it runs BEFORE user's WithActors
+                // This ensures TestProbe is available for any actors that depend on IRequiredActor<TestProbe>
+                builder.StartActors((actorSystem, actorRegistry) =>
```
Aaronontheweb, line 122: We register the `TestProbe` now BEFORE any of the user-defined actors get created, so this should prevent `ActorRegistry` deadlocks from occurring, which was happening in the `DeadlockSpec`

Aaronontheweb (commented):
Detailed my changes

Arkatufus (approved):
LGTM

PR #658, opened 2025-09-23
Feature/Add dependency injection support for health checks

beminee:
Fixes #N/A

## Changes

This PR adds dependency injection support for health checks, enabling seamless registration of health checks that require constructor dependencies like `ILogger<T>`, dbcontexts or any other services.

### New API Methods
- `WithHealthCheck<T>(string name, HealthStatus?, IEnumerable<string>?, TimeSpan?)` - Simple DI-resolved health check registration
- `WithHealthCheck<T>(AkkaHealthCheckRegistration)` - Template based registration with custom configuration

### Implementation Details
- **`DiResolvedHealthCheck<T>`**: Internal wrapper that resolves health checks from DI at runtime
- **New ctor**: Added `AkkaHealthCheckRegistration(string, HealthStatus?, IEnumerable<string>?)` for template usage

### Key Features
- Dependency injection support for health check ctors
- Automatic fallback to manual instance creation with reflection when needed

### Usage Example
```csharp
// Before: Manual injection
.WithHealthCheck(new AkkaHealthCheckRegistration("test", 
    new TestHealth(logger), // Manually supplied parameters
    HealthStatus.Unhealthy, new[] { "test" }))

// After: DI integration
.WithHealthCheck<TestHealth>("test", HealthStatus.Unhealthy, new[] { "test" })

// Or using the template
.WithHealthCheck<TestHealth>(new AkkaHealthCheckRegistration("test", 
    HealthStatus.Unhealthy, new[] { "test" }))
```

## Checklist

### DIFF
--- a/src/Akka.Hosting/AkkaConfigurationBuilder.cs
+++ b/src/Akka.Hosting/AkkaConfigurationBuilder.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Collections.Immutable;
 using System.Linq;
 using System.Reflection;
+using System.Threading;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Actor.Setup;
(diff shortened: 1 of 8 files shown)


src/Akka.Hosting/AkkaConfigurationBuilder.cs
```diff
@@ -81,6 +82,64 @@ public SerializerRegistration(string id, ImmutableHashSet<Type> typeBindings,
...
+            if (healthCheck == null)
+            {
+                healthCheck = CreateHealthCheckInstance<T>(diResolver.Resolver);
+            }
+            
+            // Delegate to the resolved health check
+            return await healthCheck.CheckHealthAsync(context, cancellationToken);
+        }
+        
+        private static T CreateHealthCheckInstance<THealthCheck>(IDependencyResolver serviceProvider) where THealthCheck : class, IAkkaHealthCheck
```
Aaronontheweb: I think if the DI registration is not available, we should hard-fail instead of having "clever" code like this to make up for it. You either did your DI registrations correctly or you didn't - no in-between.

Aaronontheweb: Side note: 

So the way we handle this inside Akka.DependendencyInjection for `Props` is

https://github.com/akkadotnet/akka.net/blob/49be5132f24bfc46219bb0cfc1a0fcd4e79d2a0a/src/contrib/dependencyinjection/Akka.DependencyInjection/ServiceProvider.cs#L96-L119

Using the `ActivatorUtilities.CreateInstance` from the Microsoft.Extensions.DI space. I think I prefer that over having hand-rolled reflection magic here in this instance.

Aaronontheweb (changes requested):
I'm not sure about the code where we're checking for DI registrations and then inventing one if it doesn't exist

src/Akka.Hosting/AkkaConfigurationBuilder.cs
beminee: @Aaronontheweb removed the DI registration magic. Now, the API should just try to resolve services and throw exception if it can't. The error will also show up on health page, i.e

```json
  {
    "name": "test",
    "status": "Unhealthy",
    "duration": "00:00:00.0100158",
    "description": "Health check type \u0027TestHealth\u0027 is not registered in the dependency injection container. Register the health check type using services.AddTransient\u003CTestHealth\u003E() or similar in your DI configuration.",
    "tags": [
      "test",
      "custom",
      "akka"
    ],
    "data": {}
  }

Aaronontheweb: I think I'm going to submit a patch to Akka.DependencyInjection to make it possible to resolve _all of the dependent services_ that the health check needs, so that way the health check itself doesn't necessarily need to be registered to the `IServiceCollection`. I'm going to do some more digging on this.

Aaronontheweb:
merged via #659

PR #667, opened 2025-10-02
Fix journal health check registration without event adapters

### LINKED ISSUE #666: Journal health checks not registered when no event adapters are configured
## Description

Journal health checks are not registered when using `.WithHealthCheck()` without adding event adapters. This is due to an early return in `AkkaPersistenceJournalBuilder.Build()` that exits before health check registration occurs.

## Reproduction

```csharp
{
    var journalOptions = new SqlJournalOptions(true)
    {
        ConnectionString = "...",
        ProviderName = ProviderName.SQLite
    };

    // ❌ Health check NOT registered - no event adapters added
    builder.WithJournal(
        journalOptions,
        journal => journal.WithHealthCheck());
}

// Result: No journal health check appears in HealthCheckService
```

## Expected Behavior

Health checks should be registered regardless of whether event adapters are configured. These are independent features.

## Actual Behavior

Health check is silently ignored. Only appears if you also add an event adapter:

```csharp
// ✅ Health check IS registered - event adapter added
builder.WithJournal(
    journalOptions,
    journal => journal
        .AddWriteEventAdapter<MyAdapter>("adapter", [typeof(MyEvent)])
        .WithHealthCheck());
```

## Root Cause

**File**: `src/Akka.Persistence.Hosting/AkkaPersistenceHostingExtensions.cs`
**Method**: `AkkaPersistenceJournalBuilder.Build()` (lines 113-133)

```csharp
internal void Build()
{
    // useless configuration - don't bother.
    if (Adapters.Count == 0 || Bindings.Count == 0)
        return;  // ← EXITS EARLY, skips health check registration!

    var adapters = new StringBuilder()
        .Append($"akka.persistence.journal.{JournalId}").Append("{");

    AppendAdapters(adapters);

    adapters.AppendLine("}");

    var finalHocon = ConfigurationFactory.ParseString(adapters.ToString())
        .WithFallback(Persistence.DefaultConfig());
    Builder.AddHocon(finalHocon, HoconAddMode.Prepend);
    
    // add the health checks if specified
    if(HealthCheckRegistration != null)
        Builder.WithHealthCheck(HealthCheckRegistration);  // ← NEVER REACHED!
}
```

The early return at line 116 exits before health check registration at line 132.

## Proposed Fix

Move health check registration **before** the early return:

```csharp
internal void Build()
{
    // add the health checks if specified - do this FIRST
    if(HealthCheckRegistration != null)
        Builder.WithHealthCheck(HealthCheckRegistration);
    
    // useless configuration - don't bother with adapters if none configured
    if (Adapters.Count == 0 || Bindings.Count == 0)
        return;
    
    var adapters = new StringBuilder()
        .Append($"akka.persistence.journal.{JournalId}").Append("{");

    AppendAdapters(adapters);

    adapters.AppendLine("}");

    var finalHocon = ConfigurationFactory.ParseString(adapters.ToString())
        .WithFallback(Persistence.DefaultConfig());
    Builder.AddHocon(finalHocon, HoconAddMode.Prepend);
}
```

This ensures health checks are registered even when no event adapters are configured.

## Impact

**Medium-High**: Users expecting health checks to work will find they're silently not registered unless they also happen to configure event adapters. This creates a confusing and non-obvious dependency between two unrelated features.

## Workaround

Always add at least one event adapter (even a no-op) when using health checks:

```csharp
builder.WithJournal(
    journalOptions,
    journal => journal
        .AddWriteEventAdapter<IdentityAdapter>("noop", [])  // Workaround
        .WithHealthCheck());
```

## Note on Snapshot Stores

This bug does NOT affect snapshot stores because `AkkaPersistenceSnapshotBuilder.Build()` doesn't have the same early return pattern - it always registers health checks.

## Discovery

This issue was discovered while refactoring `Akka.Persistence.Sql.Hosting` to use the unified API in PR https://github.com/akkadotnet/Akka.Persistence.Sql/pull/549. The refactoring exposed this edge case where health checks are registered for snapshot stores but not journals when no event adapters are present.

## Test Coverage

The existing test `JournalAndSnapshotWithBuildersSpec` (UnifiedApiSpecs.cs:224-286) doesn't catch this bug because it adds an event adapter before calling `.WithHealthCheck()`:

```csharp
journal => journal
    .AddWriteEventAdapter<UnifiedApiTestResources.TestWriteAdapter>("adapter",
        [typeof(UnifiedApiTestResources.TestEvent)])
    .WithHealthCheck(),  // Works because adapter was added first
```

**Recommended test**: Add a test that calls `.WithHealthCheck()` WITHOUT adding event adapters to ensure health checks work independently.

Aaronontheweb:
Fixes #666

## Summary
Journal health checks were not being registered when using `.WithHealthCheck()` without adding event adapters. This was caused by an early return in `AkkaPersistenceJournalBuilder.Build()` that exited before health check registration occurred.

## Changes
- Moved health check registration before the early return in `AkkaPersistenceHostingExtensions.cs:113-133` to ensure health checks are registered regardless of event adapter configuration
- Added regression test `JournalHealthCheckWithoutAdaptersSpec` that verifies health checks work without event adapters

## Testing
All 16 persistence hosting tests pass, including the new regression test that specifically tests the bug scenario.

### DIFF
--- a/src/Akka.Persistence.Hosting.Tests/UnifiedApiSpecs.cs
+++ b/src/Akka.Persistence.Hosting.Tests/UnifiedApiSpecs.cs
@@ -286,6 +286,56 @@ public sealed class JournalAndSnapshotWithBuildersSpec : Akka.Hosting.TestKit.Te
 }
 
 /// <summary>
+/// Regression test for https://github.com/akkadotnet/Akka.Hosting/issues/666
+/// Ensures journal health checks are registered even without event adapters
+/// </summary>
+public sealed class JournalHealthCheckWithoutAdaptersSpec : Akka.Hosting.TestKit.TestKit
+{
+    public JournalHealthCheckWithoutAdaptersSpec(ITestOutputHelper output) : base(output: output)
+    {
+    }
+
+    protected override void ConfigureServices(HostBuilderContext context, IServiceCollection services)
+    {
+        base.ConfigureServices(context, services);
+        services.AddHealthChecks();
+    }
+
+    protected override void ConfigureAkka(AkkaConfigurationBuilder builder, IServiceProvider provider)
+    {
+        var journalOptions = new UnifiedApiTestResources.TestJournalOptions { IsDefaultPlugin = true };
+
+        // Configure journal with health check but WITHOUT any event adapters
+        // This is the regression case from issue #666
+        builder.WithJournal(
+            journalOptions,
+            journal => journal.WithHealthCheck());
+    }
+
+    [Fact]
+    public async Task Journal_health_check_should_be_registered_without_event_adapters()
+    {
+        // assert - journal plugin configured
+        var config = Sys.Settings.Config;
+        config.HasPath("akka.persistence.journal.test-journal")
+            .Should().BeTrue();
+
+        // assert - health check is registered even without event adapters
+        var healthCheckService = Host.Services.GetRequiredService<HealthCheckService>();
+        var result = await healthCheckService.CheckHealthAsync();
+
+        result.Status.Should().Be(HealthStatus.Healthy,
+            "journal health check should be healthy");
+
+        // Verify journal health check is present
+        var journalCheck = result.Entries.FirstOrDefault(e => e.Key.Contains("test-journal"));
+        journalCheck.Should().NotBeNull("journal health check should be registered even without event adapters");
+        journalCheck.Value.Status.Should().Be(HealthStatus.Healthy,
+            "journal health check should return healthy status");
+    }
+}
+
+/// <summary>
 /// Test null builder actions work correctly
 /// </summary>
 public sealed class NullBuilderActionsSpec : Akka.Hosting.TestKit.TestKit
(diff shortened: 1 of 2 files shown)


src/Akka.Persistence.Hosting/AkkaPersistenceHostingExtensions.cs
```diff
@@ -112,6 +112,10 @@ private AkkaHealthCheckRegistration AddHealthCheck(string? name, HealthStatus un
         /// </summary>
         internal void Build()
         {
+            // add the health checks if specified - do this FIRST before any early returns
+            if(HealthCheckRegistration != null)
+                Builder.WithHealthCheck(HealthCheckRegistration);
+
             // useless configuration - don't bother.
             if (Adapters.Count == 0 || Bindings.Count == 0)
```
Aaronontheweb, line 120: This was the source of the bug.

src/Akka.Persistence.Hosting.Tests/UnifiedApiSpecs.cs
```diff
@@ -285,6 +285,56 @@ public async Task WithJournalAndSnapshot_with_builders_should_configure_everythi
     }
 }
 
+/// <summary>
+/// Regression test for https://github.com/akkadotnet/Akka.Hosting/issues/666
+/// Ensures journal health checks are registered even without event adapters
+/// </summary>
+public sealed class JournalHealthCheckWithoutAdaptersSpec : Akka.Hosting.TestKit.TestKit
```
Aaronontheweb, line 292: Regression test for #666

Aaronontheweb (commented):
Detailed my changes

Issue #665, opened 2025-10-02
Deprecate JournalOptions.Adapters property in favor of unified callback API


Aaronontheweb:
## Summary

`JournalOptions.Adapters` property should be deprecated in favor of the callback pattern introduced in PR#662. Having two competing ways to configure event adapters is confusing and inconsistent with the health check API.

## Problem

Currently there are **two ways** to configure event adapters:

### Pattern 1: Adapters Property (Legacy)
```csharp
var journalOptions = new SqlJournalOptions(true)
{
    Adapters = new AkkaPersistenceJournalBuilder("sql", builder),
    // ... other options
};
journalOptions.Adapters.AddWriteEventAdapter<ColorFruitTagger>("tagger", [typeof(string)]);

builder.WithJournal(journalOptions);
```

### Pattern 2: Callback (Unified API from PR#662)
```csharp
var journalOptions = new SqlJournalOptions(true)
{
    // ... options only
};

builder.WithJournal(
    journalOptions,
    journal => journal.AddWriteEventAdapter<ColorFruitTagger>("tagger", [typeof(string)]));
```

## Design Flaws

1. **Competing Patterns** - No clear guidance on which pattern to use
2. **Inconsistent with Health Checks** - Health checks ONLY support the callback pattern
3. **Both Can Be Used Simultaneously** - Unclear merge semantics when both are provided
4. **Dangerous Default** - Property initialized with dummy builder: `new AkkaPersistenceJournalBuilder("", null!)` (potential NRE)
5. **Violates DRY** - Two mechanisms for the same functionality

## Code Locations

**JournalOptions.cs (line 78)**:
```csharp
public AkkaPersistenceJournalBuilder Adapters { get; set; } = new ("", null!);
```

**AkkaPersistenceHostingExtensions.cs (lines 262-271)**:
```csharp
// Line 262: Applies adapters from property
builder.AddHocon(journalOptions.ToConfig(), HoconAddMode.Prepend);

// Lines 266-271: ALSO applies adapters from callback
if (configureBuilder != null)
{
    var jBuilder = new AkkaPersistenceJournalBuilder(journalOptions.Identifier, builder);
    configureBuilder(jBuilder);
    jBuilder.Build();
}
```

## Impact on Users

**Current behavior** (when both are used):
```csharp
var journalOptions = new SqlJournalOptions(true)
{
    Adapters = new AkkaPersistenceJournalBuilder("sql", builder)
};
[...]

PR #669, opened 2025-10-02
Deprecate JournalOptions.Adapters property in favor of callback API

Aaronontheweb:
Fixes #665

## Summary
The `Adapters` property on `JournalOptions` created confusion by providing two competing patterns for configuring event adapters. This PR deprecates the property in favor of the unified callback pattern introduced in PR #662.

## Problem
Having two ways to configure event adapters was:
- ❌ Confusing - No clear guidance on which pattern to use
- ❌ Inconsistent - Health checks only support callbacks
- ❌ Undocumented - Users could use both simultaneously with unclear merge semantics
- ❌ Dangerous - Property initialized with dummy builder that could cause NRE

## Solution
1. **Mark property as `[Obsolete]`** with clear migration guidance pointing to issue #665
2. **Remove internal usage** - Stop reading the property in `JournalOptions.Build()`
3. **Add regression test** - Verify deprecated property is ignored and callbacks work

## Changes
- `JournalOptions.cs:78` - Added `[Obsolete]` attribute to `Adapters` property
- `JournalOptions.cs:105` - Removed `Adapters.AppendAdapters(sb)` call
- `EventAdapterSpecs.cs:134-207` - Added `DeprecatedAdaptersPropertySpec` regression test

## Benefits
✅ Single, consistent pattern for configuring adapters and health checks  
✅ Cleaner API surface - options contain only configuration data  
✅ Clear migration path with deprecation warning  
✅ No silent failures - deprecated property is truly ignored  

## Migration Example

**Before (deprecated):**
```csharp
var journalOptions = new SqlJournalOptions(true)
{
    Adapters = new AkkaPersistenceJournalBuilder("sql", builder)
};
journalOptions.Adapters.AddWriteEventAdapter<Foo>("foo", [typeof(Bar)]);

builder.WithJournal(journalOptions);
```

**After (recommended):**
```csharp
var journalOptions = new SqlJournalOptions(true);

builder.WithJournal(
    journalOptions,
    journal => journal.AddWriteEventAdapter<Foo>("foo", [typeof(Bar)]));
```

## Testing
All 17 persistence hosting tests pass, including new regression test that verifies:
- Deprecated property is ignored (doesn't configure adapters)
- Callback pattern works correctly
- Uses `#pragma warning disable` to test deprecated API without warnings

## Note
This is Phase 1 of the deprecation plan. The property will be removed entirely in v1.6.0.

Issue #552, opened 2025-01-13
Phobos instrumentation settings for actor props get overwritten by PhobosActorCell settings


object:
I continue struggling with custom actor type name, and I believe this time I hit an issue that may be unrelated to F#.

I debug the code that creates an instance of an actor by calling ActorOf on an actor factory. If I debug the code, I reach the following Phobos code:

public override IActorRef ActorOf(Props props, string name = null)
    {
      return base.ActorOf(this.InstrumentationSettings.PropagateSettingsToChildren ? props.WithInstrumentation(this.InstrumentationSettings) : props, name);
    }

The custom actor type name is inside "props" data that are sent to the method ActorOf. But the flag PropagateSettingsToChildren on the instance of PhobosActorCell is set to "true", so instead of using the "props" data that are sent to the method it uses data that belong to this.InstrumentationSettings, and there ActorTypeName is set to "null". So whatever I send gets overwritten by the InstrumentationSettings of the PhobosActorCell.

I never set the flag PropagateSettingsToChildren.

So how should I trigger usage of setting from my props, and not the PhobosActorCell?

object:
Oops, that is probably wrong repo for this issue. I will open one in Phobos.

object:
Closing this, created https://github.com/petabridge/phobos-issues/issues/85

PR #674, opened 2025-10-13
Fix event adapter callback API not invoking adapters at runtime

Aaronontheweb:
## Summary

Fixed a critical bug where event adapters configured via the NEW callback API were not being invoked at runtime, despite being correctly present in HOCON configuration.

## Root Cause

`AkkaPersistenceJournalBuilder.Build()` was creating a Config with `.WithFallback(Persistence.DefaultConfig())` which interfered with the main journal configuration during HOCON merging, causing adapters to be registered in config but not invoked during persistence operations.

**File**: `src/Akka.Persistence.Hosting/AkkaPersistenceHostingExtensions.cs` line 130

## The Fix

Removed the unnecessary `.WithFallback(Persistence.DefaultConfig())` call. The adapter configuration blocks (`event-adapters` and `event-adapter-bindings`) don't require fallback to default persistence config and merge correctly with the journal config without it.

## Impact

This bug affected users with:
- Multiple journal configurations (e.g., default + sharding journals)
- NEW callback API: `journalBuilder: journal => journal.AddWriteEventAdapter<...>()`
- Explicit `JournalOptions` + `WithJournalAndSnapshot` pattern (common in Akka.Persistence.Sql)

**This fix is especially critical** now that the old `Adapters` property has been deprecated in #665. All users must now use the callback API which was broken without this fix.

## Testing

- Added comprehensive regression test suite (`EventAdapterRuntimeInvocationSpecs.cs`)
  - Tests verify adapter runtime invocation with various configuration patterns
  - Covers simple, multi-journal, and explicit JournalOptions scenarios
- All 20 `Akka.Persistence.Hosting.Tests` pass ✅
- Verified backwards compatibility with existing `EventAdapterSpecs` ✅
- Tested with Akka.Persistence.Sql - resolves reported issue ✅

### Before Fix
```
Event adapter was called 0 times
Expected: 3, Actual: 0
```

### After Fix
```
[TestEventTagger] ToJournal called 1 times. Event: TestEvent, Result: Tagged
[TestEventTagger] ToJournal called 2 times. Event: TestEvent, Result: Tagged
[TestEventTagger] ToJournal called 3 times. Event: TestEvent, Result: Tagged
Test Run Successful.
```

## Related Issues

- Fixes runtime invocation issue reported in https://github.com/akkadotnet/Akka.Persistence.Sql/issues/552
- Critical fix for #665 callback API requirement

## Changes

- Modified: `src/Akka.Persistence.Hosting/AkkaPersistenceHostingExtensions.cs` (removed 1 line)
- Added: `src/Akka.Persistence.Hosting.Tests/EventAdapterRuntimeInvocationSpecs.cs` (254 lines, 3 test classes)

### DIFF
--- a/src/Akka.Persistence.Hosting.Tests/EventAdapterRuntimeInvocationSpecs.cs
+++ b/src/Akka.Persistence.Hosting.Tests/EventAdapterRuntimeInvocationSpecs.cs
@@ -1,11 +1,15 @@
 using System;
 using System.Collections.Generic;
-using System.Threading;
+using System.Linq;
 using System.Threading.Tasks;
 using Akka.Actor;
 using Akka.Configuration;
 using Akka.Hosting;
 using Akka.Persistence.Journal;
+using Akka.Persistence.Query;
+using Akka.Persistence.Query.InMemory;
+using Akka.Streams;
+using Akka.Streams.Dsl;
 using FluentAssertions;
 using Xunit;
 using Xunit.Abstractions;
@@ -13,9 +17,9 @@ using Xunit.Abstractions;
 namespace Akka.Persistence.Hosting.Tests;
 
 /// <summary>
-/// Tests that verify event adapters are actually INVOKED at runtime when using the NEW callback API.
-/// This is a regression test for https://github.com/akkadotnet/Akka.Persistence.Sql/issues/552
-/// where event adapters were present in HOCON but never invoked at runtime.
+/// Regression test for https://github.com/akkadotnet/Akka.Persistence.Sql/issues/552
+/// Verifies that event adapters configured via the NEW callback API are actually invoked at runtime
+/// by checking that events are tagged and appear in EventsByTag queries.
 /// </summary>
 public class EventAdapterRuntimeInvocationSpecs : Akka.Hosting.TestKit.TestKit
 {
@@ -34,17 +38,17 @@ public class EventAdapterRuntimeInvocationSpecs : Akka.Hosting.TestKit.TestKit
         public string Data { get; }
     }
 
-    public sealed class InvocationCountingAdapter : IWriteEventAdapter
+    /// <summary>
+    /// Event adapter that tags TestEvent instances
+    /// </summary>
+    public sealed class TestEventTagger : IWriteEventAdapter
     {
-        public static int CallCount = 0;
-
-        public InvocationCountingAdapter(ExtendedActorSystem system) { }
+        public TestEventTagger(ExtendedActorSystem system) { }
 
         public string Manifest(object evt) => string.Empty;
 
         public object ToJournal(object evt)
         {
-            Interlocked.Increment(ref CallCount);
             return evt switch
             {
                 TestEvent => new Tagged(evt, new[] { "test-tag" }),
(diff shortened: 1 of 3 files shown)


Aaronontheweb:
Updated tests to use `AwaitAssertAsync` instead of `Task.Delay` + immediate assertions. This provides proper retry logic and eliminates arbitrary wait times.

src/Akka.Persistence.Hosting.Tests/EventAdapterRuntimeInvocationSpecs.cs
```diff
@@ -0,0 +1,543 @@
...
+using Xunit.Abstractions;
+
+namespace Akka.Persistence.Hosting.Tests;
+
+/// <summary>
+/// Tests that verify event adapters are actually INVOKED at runtime when using the NEW callback API.
+/// This is a regression test for https://github.com/akkadotnet/Akka.Persistence.Sql/issues/552
+/// where event adapters were present in HOCON but never invoked at runtime.
+/// </summary>
+public class EventAdapterRuntimeInvocationSpecs : Akka.Hosting.TestKit.TestKit
```
Aaronontheweb, line 24: Reproduces  https://github.com/akkadotnet/Akka.Persistence.Sql/issues/552 using the in-memory journal.

src/Akka.Persistence.Hosting.Tests/EventAdapterRuntimeInvocationSpecs.cs
```diff
@@ -0,0 +1,543 @@
...
+
+    public sealed class TestEvent
+    {
+        public TestEvent(string data) { Data = data; }
+        public string Data { get; }
+    }
+
+    public sealed class InvocationCountingAdapter : IWriteEventAdapter
+    {
+        public static int CallCount = 0;
```
Aaronontheweb: Static fields shouldn't / won't work if there are multiple tests all running in the same fixture. Probably should come up with something else here, like an Akka.Persistence.Query replay.

src/Akka.Persistence.Hosting/AkkaPersistenceHostingExtensions.cs
```diff
@@ -127,8 +127,7 @@ internal void Build()
 
             adapters.AppendLine("}");
 
-            var finalHocon = ConfigurationFactory.ParseString(adapters.ToString())
-                .WithFallback(Persistence.DefaultConfig()); // add the default config as a fallback
```
Aaronontheweb, line 131: This is the real bug - the default HOCON getting prepended to the front of the configuration causes the event-adapters collection to be empty.

Aaronontheweb (commented):
Detailed my changes - I want to improve the tests but the substantial changes are good IMHO

Aaronontheweb:
## Verification Complete ✅

Tested this fix using a local NuGet build (v1.5.52-alpha-1760450556) against the reproduction tests in https://github.com/akkadotnet/Akka.Persistence.Sql/pull/554.

### Test Results

**NEW Callback API (the affected API from issue #552):**
- ✅ `RuntimeEventAdapterSpec.EventAdapter_ShouldWork_OnDefaultJournal` - **PASSES**
- ✅ `RuntimeEventAdapterSpec.EventAdapter_ShouldWork_OnShardingJournal_ReproducesUserScenario` - **PASSES**

Both tests that use the new callback API pattern now pass with this fix. The sharding journal test specifically reproduces the exact scenario from https://github.com/akkadotnet/Akka.Persistence.Sql/issues/552 where event adapters configured via `journalBuilder:` parameter were not being invoked.

**OLD Deprecated API:**
- ❌ `RuntimeEventAdapterOldApiSpec.EventAdapter_ShouldWork_WithOldApi` - **FAILS (Expected)**

The old `JournalOptions.Adapters` property API was intentionally disabled in #669, so this failure is expected and correct.

### Conclusion

This fix successfully resolves https://github.com/akkadotnet/Akka.Persistence.Sql/issues/552. Event adapters configured via the new callback API now invoke correctly at runtime in multi-journal configurations.

The root cause fix (removing the `.WithFallback(Persistence.DefaultConfig())` call) was exactly right - it was interfering with HOCON merging and preventing adapter registration from taking effect.

Arkatufus (approved):
LGTM

PR #706, opened 2026-01-26
feat: add OpenTelemetry trace correlation for LoggerFactoryLogger

### LINKED ISSUE #700: Add OpenTelemetry trace correlation support to LoggerFactoryLogger
## Summary

Update `LoggerFactoryLogger` to leverage the new `LogEvent.ActivityContext` property (akkadotnet/akka.net#7995) to enable log/trace correlation with OpenTelemetry.

## Background

Related to akkadotnet/akka.net#6855 - `Activity.Current` uses `AsyncLocal<T>` which doesn't flow across actor mailbox boundaries. When actors log messages, `Activity.Current` is null by the time the log event is processed by the logger actor.

akkadotnet/akka.net#7995 adds `ActivityContext?` capture to `LogEvent` at creation time, preserving the trace context before the mailbox crossing.

### 1. Add `AkkaLogState` struct

A struct implementing `IReadOnlyList<KeyValuePair<string, object?>>` that passes trace context as structured log state:

```csharp
public readonly struct AkkaLogState : IReadOnlyList<KeyValuePair<string, object?>>
{
    public const string TraceIdKey = "Akka.TraceId";
    public const string SpanIdKey = "Akka.SpanId";
    public const string TraceFlagsKey = "Akka.TraceFlags";

    // Implementation passes TraceId, SpanId, TraceFlags as string attributes
}
```

### 2. Add `AkkaTraceContextProcessor`

A `BaseProcessor<LogRecord>` that extracts trace context from log attributes and sets `LogRecord.TraceId/SpanId` directly:

```csharp
public sealed class AkkaTraceContextProcessor : BaseProcessor<LogRecord>
{
    public override void OnEnd(LogRecord logRecord)
    {
        // Skip if already has trace context
        if (logRecord.TraceId != default) return;
        
        // Extract from attributes and set LogRecord.TraceId/SpanId directly
    }
}
```

### 3. Update `LoggerFactoryLogger.Log()`

When `LogEvent.ActivityContext` has a value, use `AkkaLogState` to pass the trace context:

```csharp
protected virtual void Log(LogEvent log, ActorPath path)
{
    if (log.ActivityContext.HasValue)
    {
        var ctx = log.ActivityContext.Value;
        var state = new AkkaLogState(ctx.TraceId, ctx.SpanId, ctx.TraceFlags, log.Message.ToString());
        _logger.Log(GetLogLevel(log), new EventId(), state, log.Cause, (s, _) => s.ToString());
    }
    else
    {
        // Existing logging path
    }
}
```

### 4. Add extension method for easy configuration

```csharp
public static class AkkaOpenTelemetryExtensions
{
    public static OpenTelemetryLoggerOptions AddAkkaTraceCorrelation(
        this OpenTelemetryLoggerOptions options)
    {
        options.AddProcessor(new AkkaTraceContextProcessor());
        return options;
    }
}
```

## User Configuration

Users will need to explicitly register the processor in their OpenTelemetry configuration:

```csharp
builder.Logging.AddOpenTelemetry(options =>
{
    options.AddAkkaTraceCorrelation();  // Extension method from Akka.Hosting
    options.AddOtlpExporter();
});
```

**Note:** This cannot be automatically registered from `AddAkka()` because logging is configured via `ILoggingBuilder` before Akka.Hosting runs, and OpenTelemetry doesn't use the `IOptions<T>` pattern for logging configuration.

## Reference Implementation

A working proof-of-concept is available at: https://github.com/Aaronontheweb/akka.net-log-trace-correlation-POC

## Dependencies

- akkadotnet/akka.net#7995 must be merged first

Aaronontheweb:
## Summary

Implements GitHub issue #700: Add OpenTelemetry trace correlation support to `LoggerFactoryLogger`.

**Problem:** `Activity.Current` (used by OpenTelemetry) doesn't flow across actor mailbox boundaries because it uses `AsyncLocal<T>`. This means logs generated by actors lose their trace context correlation.

**Solution:** Leverage `LogEvent.ActivityContext` added in Akka.NET 1.5.59, which captures trace context at log event creation time before the mailbox crossing.

## Changes

- **`AkkaLogState`** - Low-allocation struct using lazy enumeration (`yield return`) to propagate trace context in MEL state dictionaries. Stores `ActivityTraceId`/`ActivitySpanId` as structs directly to avoid `ToString()` allocations.

- **`AkkaTraceContextProcessor`** - `BaseProcessor<LogRecord>` that extracts `Akka.TraceId`, `Akka.SpanId`, `Akka.TraceFlags` from log attributes and sets them on `LogRecord` for OpenTelemetry exporters.

- **`AddAkkaTraceCorrelation()`** - Extension method on `OpenTelemetryLoggerOptions` for easy configuration.

- **`LoggerFactoryLogger`** - Updated to include `ActivityContext` in log state when available.

- **Dependencies** - Bumped Akka.NET to 1.5.59, Microsoft.Extensions to 8.0.0

- **Aspire Demo** - Added demo project (`src/Examples/Akka.Hosting.OpenTelemetry.Demo/`) with Seq for visual validation

## Usage

```csharp
builder.Logging.AddOpenTelemetry(options =>
{
    options.SetResourceBuilder(ResourceBuilder.CreateDefault()
        .AddService("my-service"));

    // Register processor FIRST (before exporters)
    options.AddAkkaTraceCorrelation();

    options.AddOtlpExporter();
});

services.AddAkka("MySystem", configBuilder =>
{
    configBuilder.ConfigureLoggers(setup =>
    {
        setup.ClearLoggers();
        setup.AddLoggerFactory();
    });
});
```

## Test plan

- [x] Unit tests for `AkkaLogState` (8 tests)
- [x] Unit tests for `AkkaTraceContextProcessor` (10 tests)
- [x] All existing tests pass (297 tests)
- [ ] Manual validation with Aspire demo

Closes #700

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,11 +2,12 @@
   <PropertyGroup>
     <Copyright>Copyright © 2013-$([System.DateTime]::Now.Year) Akka.NET Team</Copyright>
     <Authors>Akka.NET Team</Authors>
-    <VersionPrefix>1.5.57</VersionPrefix>
+    <VersionPrefix>1.5.59</VersionPrefix>
     <PackageReleaseNotes>**New Features**
-* [Add semantic logging support for Akka.NET 1.5.56+](https://github.com/akkadotnet/Akka.Hosting/pull/693) - enables Microsoft.Extensions.Logging to receive properly structured state dictionaries instead of pre-formatted strings. When using Akka.NET 1.5.56+, log messages now include structured properties from the semantic logging API along with Akka metadata (ActorPath, Timestamp, Thread, LogSource). Fully backwards compatible with older Akka.NET versions.
+* [Add OpenTelemetry trace correlation support for LoggerFactoryLogger](https://github.com/akkadotnet/Akka.Hosting/issues/700) - enables proper trace correlation for logs emitted from actor code. Solves the problem that Activity.Current doesn't flow across actor mailbox boundaries. When using Akka.NET 1.5.59+, LogEvent.ActivityContext captures trace context at log creation time and flows it through to OpenTelemetry LogRecords via the new AkkaTraceContextProcessor.
 **Updates**
-* [Bump Akka version from 1.5.55 to 1.5.57](https://github.com/akkadotnet/akka.net/releases/tag/1.5.57)</PackageReleaseNotes>
+* [Bump Akka version from 1.5.57 to 1.5.59](https://github.com/akkadotnet/akka.net/releases/tag/1.5.59)
+* Added OpenTelemetry dependency (1.9.0+) for trace correlation support</PackageReleaseNotes>
     <PackageIcon>akkalogo.png</PackageIcon>
     <PackageProjectUrl>
       https://github.com/akkadotnet/Akka.Hosting
@@ -28,9 +29,9 @@
     <TestSdkVersion>17.11.1</TestSdkVersion>
     <CoverletVersion>6.0.3</CoverletVersion>
     <XunitRunneVisualstudio>3.1.5</XunitRunneVisualstudio>
-    <AkkaVersion>1.5.57</AkkaVersion>
-    <MicrosoftExtensionsVersion>[6.0.0,)</MicrosoftExtensionsVersion>
-    <SystemTextJsonVersion>[6.0.10,)</SystemTextJsonVersion>
+    <AkkaVersion>1.5.59</AkkaVersion>
+    <MicrosoftExtensionsVersion>[8.0.0,)</MicrosoftExtensionsVersion>
+    <SystemTextJsonVersion>[8.0.5,)</SystemTextJsonVersion>
   </PropertyGroup>
   <!-- SourceLink support for all Akka.NET projects -->
   <ItemGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -35,8 +35,8 @@
     <CoverletVersion>6.0.3</CoverletVersion>
     <XunitRunneVisualstudio>3.1.5</XunitRunneVisualstudio>
     <AkkaVersion>1.5.59</AkkaVersion>
-    <MicrosoftExtensionsVersion>[6.0.0,)</MicrosoftExtensionsVersion>
-    <SystemTextJsonVersion>[6.0.10,)</SystemTextJsonVersion>
+    <MicrosoftExtensionsVersion>[8.0.0,)</MicrosoftExtensionsVersion>
+    <SystemTextJsonVersion>[8.0.5,)</SystemTextJsonVersion>
   </PropertyGroup>
   <!-- SourceLink support for all Akka.NET projects -->
   <ItemGroup>
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -35,8 +35,8 @@
     <CoverletVersion>6.0.3</CoverletVersion>
     <XunitRunneVisualstudio>3.1.5</XunitRunneVisualstudio>
     <AkkaVersion>1.5.59</AkkaVersion>
-    <MicrosoftExtensionsVersion>[8.0.0,)</MicrosoftExtensionsVersion>
-    <SystemTextJsonVersion>[8.0.5,)</SystemTextJsonVersion>
+    <MicrosoftExtensionsVersion>[6.0.0,)</MicrosoftExtensionsVersion>
+    <SystemTextJsonVersion>[6.0.10,)</SystemTextJsonVersion>
   </PropertyGroup>
   <!-- SourceLink support for all Akka.NET projects -->
   <ItemGroup>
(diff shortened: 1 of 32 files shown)


Aaronontheweb:
Validated this approach works for Phobos log/trace correlation.

We created a prototype test using the `AkkaLogState` and `AkkaTraceContextProcessor` implementations from this PR to verify that trace context correctly propagates through the MEL state dictionary. The test confirms that `TraceId`, `SpanId`, and `TraceFlags` are all properly captured and accessible.

This will allow us to deprecate Phobos's `AppendLogsToTrace` feature (which was a workaround for this exact problem) and point users to `LoggerFactoryLogger` + `AddAkkaTraceCorrelation()` as the proper solution.

src/Akka.Hosting/Logging/LoggerFactoryLogger.cs
```diff
@@ -64,46 +64,61 @@ protected virtual void Log(LogEvent log, ActorPath path)
...
-                    return $"Received a malformed formatted message. Log level: [{log.LogLevel()}], Message: [{log.Message}]";
+                    // Fallback for non-structured messages without trace context
+                    _akkaLogger.Log<LogEvent>(logLevel, new EventId(), log, log.Cause,
+                        (@event, exception) => formattedMessage);
                 }
-            });
+            }
+        }
+
+        private static string SafeFormat(LogEvent log)
```
Aaronontheweb, line 109: While all of these logging changes are still new, want to be careful about not err-ing out if there are formatting issues / config conflicts

Aaronontheweb:
Some screenshots from the Aspire demo:

## Traces

image


## Logs Correlated to Trace

These were all captured via the `ILoggingAdapter`

image

src/Akka.Hosting/Logging/AkkaLogState.cs
```diff
@@ -0,0 +1,151 @@
...
+    /// and yields values lazily via the enumerator rather than copying into a new collection.
+    /// </para>
+    /// <para>
+    /// This struct is used to propagate trace context (TraceId, SpanId, TraceFlags) from the
+    /// originating actor thread to the logging infrastructure, solving the problem that
+    /// <see cref="Activity.Current"/> doesn't flow across actor mailbox boundaries because
+    /// it uses <see cref="AsyncLocal{T}"/>.
+    /// </para>
+    /// </remarks>
+    public readonly struct AkkaLogState : IEnumerable<KeyValuePair<string, object?>>
```
Aaronontheweb: Common pattern we use to prevent allocations - a stack-allocated `IEnumerable` property bag. We use this in the itnternal logger implementation in Akka.NET Itself.

Aaronontheweb: Should not be public though

src/Akka.Hosting/Logging/AkkaTraceContextProcessor.cs
```diff
@@ -0,0 +1,168 @@
...
+            {
+                return TryParseSpanId(spanIdStr);
+            }
+
+            return null;
+        }
+
+        private static void SetTraceContext(LogRecord record, ActivityTraceId traceId, ActivitySpanId spanId, ActivityTraceFlags traceFlags)
+        {
+            // LogRecord has internal setters, so we need to use reflection
```
Aaronontheweb, line 114: this is true unfortunately, per the path the OTEL authors recommended to us https://github.com/open-telemetry/opentelemetry-dotnet/issues/6085

Aaronontheweb: Going to update this to use `UnsafeAccessor` instead

Aaronontheweb (commented):
Reviewed some of the changes here, have some fixes to make

src/Akka.Hosting/Logging/AkkaTraceContextProcessor.cs
Aaronontheweb: lol can't - due to .NET Framework

src/Akka.Hosting/Logging/AkkaLogState.cs
Aaronontheweb: Resolved - this is now private.

Directory.Build.props
```diff
@@ -31,8 +35,8 @@
     <CoverletVersion>6.0.3</CoverletVersion>
     <XunitRunneVisualstudio>3.1.5</XunitRunneVisualstudio>
     <AkkaVersion>1.5.59</AkkaVersion>
-    <MicrosoftExtensionsVersion>[6.0.0,)</MicrosoftExtensionsVersion>
-    <SystemTextJsonVersion>[6.0.10,)</SystemTextJsonVersion>
+    <MicrosoftExtensionsVersion>[8.0.0,)</MicrosoftExtensionsVersion>
```
Aaronontheweb, line 38: The OTEL packages require these

README.md
Aaronontheweb, line 1: Added documentation

Aaronontheweb (commented):
Detailed my changes - good to go for a beta release IMHO

Arkatufus (approved):
LGTM

PR #501, opened 2024-10-09
Avoid caching nobody when actor doesn't exist in registry

JamesNorris153:
Fixes #502 

## Changes

Avoids caching `Nobody` as the actor ref for a `RequiredActor` if it has not been registered yet

PR #718, opened 2026-03-11
Fix #717 by not caching the _internalRef when calling .TryGet on the …

### LINKED ISSUE #717: RequiredActor caches ActorRefs.Nobody in the synchronous ActorRef path if actor is not registered yet
**Version Information**
Version of Akka.NET? 1.5.57
Which Akka.NET Modules? Akka.Hosting

**Describe the bug**

Same issue as #501 but for the synchronous path. Will prepare a PR shortly (same fix).

v3xro:
…ActorRegistry

Fixes #717

## Changes

This fix applies the same issue as #501 but in a different place.


Aaronontheweb (approved):
LGTM

PR #724, opened 2026-03-30
Fix LoggerFactoryLogger metadata regression for non-semantic logs

Aaronontheweb:
## Summary

- **Fix metadata regression**: PR #706 (trace correlation) accidentally gated `ActorPath`, `LogSource`, `Timestamp`, and `Thread` emission on `_hasSemanticProperties`, which was `false` for plain-string logs. Non-semantic logs (~19% of production volume) lost all structured metadata, breaking downstream consumers (Datadog pipelines, custom ConsoleFormatters).
- **Remove dead code**: `_hasSemanticProperties` field was always `true` in both constructors after the fix, making the `else` branch in `GetEnumerator()` unreachable. Removed the field and dead branch entirely — the `_semanticProperties != null` check handles the structured/non-structured distinction.

## Test plan

- [x] Updated 2 existing `AkkaLogStateSpecs` unit tests for new constructor signature
- [x] Added `PlainStringLogsShouldIncludeMetadata` — verifies metadata present for `log.Info("plain string")`
- [x] All 19 logging-related tests pass
- [ ] Verify snapshot tests in #726 will validate the regression end-to-end


src/Akka.Hosting/Logging/AkkaLogState.cs
```diff
@@ -115,31 +122,23 @@ public AkkaLogState(ActivityContext activityContext, string formattedMessage)
                 yield return new KeyValuePair<string, object?>(TraceFlagsKey, (int)_activityContext.TraceFlags);
             }
 
-            if (_hasSemanticProperties)
```
Aaronontheweb, line 118: This is what caused the regression in #706 - this property stopped non-semantic logs from emitting semantic properties.

Aaronontheweb (commented):
LGTM

PR #725, opened 2026-03-30
Fix flaky cluster tests: startup ordering and timeout handling

Aaronontheweb:
## Summary

- **Fix CancellationToken race**: `TestHelper.CreateHost` passed a 10-second `CancellationToken` to `host.StartAsync()`. When the token fired on slow CI, the .NET Generic Host triggered `StopAsync`/`CoordinatedShutdown` while `AkkaHostedService.StartAsync` was still iterating through `_actorStarters`, causing `InvalidOperationException: Cannot create child while terminating or terminated`. Fix: start host without cancellation token; use a separate 30-second timeout only for the cluster formation wait.
- **Fix startup ordering**: Cluster join was registered via `AddStartup` (`_startupTasks`), but shard region start was registered via `WithShardRegion` (`_actorStarters`). Since `_actorStarters` always execute before `_startupTasks` in `AkkaConfigurationBuilder.StartAsync`, the shard region tried to initialize before the node joined the cluster. Fix: move cluster join to `WithActors` so it runs as an `_actorStarter` before any cluster-dependent starters.
- Applied to `TestHelper`, `ShardedDaemonProcessProxySpecs`, and `ProxySystem`.

## Test plan

- [x] `ClusterShardingSpecs.Should_use_ActorRegistry_with_ShardRegion` passes deterministically
- [x] `ShardedDaemonProcessProxySpecs.ShardedDaemonProcessProxy_must_start_daemon_process_on_proxy` passes deterministically
- [x] Full cluster test suite: 34/34 passed, 0 failed

### DIFF
--- a/src/Akka.Cluster.Hosting.Tests/ShardedDaemonProcessProxySpecs.cs
+++ b/src/Akka.Cluster.Hosting.Tests/ShardedDaemonProcessProxySpecs.cs
@@ -40,6 +40,8 @@ public class ShardedDaemonProcessProxySpecs: Akka.Hosting.TestKit.TestKit
             {
                 Roles = [Role]
             })
+            // Join cluster via WithActors (not AddStartup) so it runs before
+            // WithShardedDaemonProcess, which needs a formed cluster.
             .WithActors((system, _) =>
             {
                 var cluster = Cluster.Get(system);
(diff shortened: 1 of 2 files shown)


src/Akka.Cluster.Hosting.Tests/ShardedDaemonProcessProxySpecs.cs
```diff
@@ -40,21 +40,23 @@ protected override void ConfigureAkka(AkkaConfigurationBuilder builder, IService
             {
                 Roles = [Role]
             })
+            // Join cluster via WithActors (not AddStartup) so it runs before
+            // WithShardedDaemonProcess, which needs a formed cluster.
+            .WithActors((system, _) =>
```
Aaronontheweb, line 45: This is just a small ordering optimization

Aaronontheweb (commented):
Detailed my changes

PR #728, opened 2026-03-31
[xUnit 3] Convert Akka.Hosting.TestKit and all tests to xUnit 3

Arkatufus:
## Changes

* Convert Akka.Hosting.TestKit to use the new xUnit 3 Akka.TestKit.Xunit
* **Add Akka.Hosting.TestKit.Xunit2 that still supports the old xUnit 2**
* Convert all tests (including API approval) to use xUnit 3


Aaronontheweb:
closed via #729

PR #735, opened 2026-04-23
fix: resolve implicit-sender leak under xUnit v3 parallel execution

### LINKED ISSUE #733: `Akka.Hosting.TestKit` implicit-sender leaks across `ActorSystem`s under xUnit v3 parallel class execution
## Version Information

**Version of Akka.NET?** 1.5.65
**Version of Akka.Hosting?** 1.5.65 (regression introduced with the xUnit v3 migration in 1.5.64)
**Which Akka.NET Modules?** `Akka`, `Akka.Hosting`, `Akka.Hosting.TestKit`, `Akka.Persistence`, `Akka.TestKit.Xunit`
**Test framework:** `xunit.v3` 3.2.2, `xunit.runner.visualstudio` 3.1.5
**TFM:** `net8.0` / `net10.0` (reproduced on `net8.0`; reporter observed on `net10.0`)

## Describe the bug

`Akka.Hosting.TestKit` sets `InternalCurrentActorCellKeeper.Current = TestActor.Underlying` in several places during test initialization so that, inside a test body, the single-argument extension `actor.Tell(msg)` picks up `TestActor` as the implicit sender. That field is backed by `[ThreadStatic]`, which is per-thread — **not** per async context.

Under xUnit v3's default parallel-class execution, test classes run concurrently on the shared `ThreadPool`. An `await` continuation inside test *A* can resume on a thread whose `Current` was just written by test *B*'s init code. Test *A*'s `.Tell(msg)` then dispatches with `Sender = B's TestActor` — an `IActorRef` belonging to a **different `ActorSystem`**. When the receiving actor does `Sender.Tell(reply)`, the reply crosses `ActorSystem` boundaries and lands in the wrong `TestActor`'s mailbox.

This silently produces one of two symptoms:

- `ExpectMsg<T>` times out in the test that sent the message (its reply went elsewhere).
- `ExpectNoMsg` / `ExpectTerminated` fails in a sibling test that receives the stray reply from a foreign `ActorSystem`.

It did not appear before `Akka.Hosting.TestKit` 1.5.64 because the xUnit v2 default was to run test classes in an assembly serially; xUnit v3 flipped that default to parallel.

## To Reproduce

Minimal repro (adapted from a report against `akkadotnet/Akka.Hosting` `dev`):

```csharp
using Akka.Actor;
using Akka.Hosting;
using Akka.Persistence;
using Akka.Persistence.Hosting;
using Xunit;

public record CreateOrder(int Id);
public record OrderCommandResponse(int Id);
public record OrderCreatedEvent(int Id);

public class OrderActor : ReceivePersistentActor
{
    public OrderActor(int id)
    {
        PersistenceId = "order-" + id;
        Command<CreateOrder>(cmd => Persist(new OrderCreatedEvent(cmd.Id), _ =>
        {
            Sender.Tell(new OrderCommandResponse(cmd.Id));
        }));
        Recover<OrderCreatedEvent>(_ => { });
    }
    public override string PersistenceId { get; }
}

public abstract class ReproKit : Akka.Hosting.TestKit.TestKit
{
    protected ReproKit() : base(actorSystemName: "test-" + Guid.NewGuid().ToString("N")) { }

        => builder.WithInMemoryJournal().WithInMemorySnapshotStore();

    protected async Task BuggyPattern()
    {
        var id = Random.Shared.Next(1, 100);
        var actor = Sys.ActorOf(Props.Create(() => new OrderActor(id)), "order-" + id);

        await Task.Yield();
        await Task.Delay(Random.Shared.Next(5, 30));

        actor.Tell(new CreateOrder(id));   // implicit sender — resolved from [ThreadStatic] Current
        try { await ExpectMsgAsync<OrderCommandResponse>(TimeSpan.FromSeconds(3)); } catch { }
    }

    protected async Task SilentPattern()
    {
        await Task.Yield();
        await Task.Delay(Random.Shared.Next(5, 30));
        await ExpectNoMsgAsync(TimeSpan.FromMilliseconds(300));
    }
}

public class B01 : ReproKit { [Fact] public Task Bug() => BuggyPattern(); }
// ...B02–B10 identical...
public class S01 : ReproKit { [Fact] public Task S() => SilentPattern(); }
// ...S02–S05 identical...
```

`xunit.runner.json` must enable parallel class execution:

```json
{ "parallelizeAssembly": false, "parallelizeTestCollections": true }
```

Then:

```
dotnet test -c Release
```

## Expected behavior

Each test's `actor.Tell(msg)` should resolve the implicit sender to **that test's** `TestActor`. No message should ever cross `ActorSystem` boundaries unless user code explicitly routes it there.

## Actual behavior

Sibling tests receive replies that originated in a different `ActorSystem`. Typical failure output:

```
Akka.Hosting.Tests.ImplicitSenderLeak.S03.S FAIL
  Expected no messages during 00:00:00.3000000, instead we received
    <OrderCommandResponse { Id = 14 }>
    from [akka://test-11ed15ad25ea4392b2c3820f6ffac35c/user/order-14]
```

`S03` has its own `ActorSystem`, never creates any `OrderActor`, and only calls `ExpectNoMsg` — yet its `TestActor` received a reply from an actor in a foreign `ActorSystem`. The only explanation is that a sibling test's `.Tell(new CreateOrder(id))` dispatched with `Sender = S03.TestActor`.

### Reproducibility measured locally

Adding the 15-class repro above to `Akka.Persistence.Hosting.Tests` (xUnit v3) with `parallelizeTestCollections: true`, and running `dotnet test -c Release` five times in a row on an 8-core Windows box:

| Run | Outcome |
|-----|---------|
| 1   | **FAIL** — 1 stray cross-system delivery |
| 2   | **FAIL** — 1 stray cross-system delivery |
| 3   | **FAIL** — 3 stray cross-system deliveries |
| 4   | **FAIL** — 2 stray cross-system deliveries |
| 5   | PASS    |

Consistent with the original report of "at least one failure in 50/50 runs" on a larger test matrix.

## Root cause

Three pre-existing design choices line up for the first time under xUnit v3:

1. **`akka.net/src/core/Akka/Actor/Internal/InternalCurrentActorCellKeeper.cs:18`**
   ```csharp
   [ThreadStatic] private static ActorCell? _current;
   ```
   Per-thread storage, does not flow across `await`.

2. **`Akka.Hosting/src/Akka.Hosting.TestKit/TestKit.Shared.cs`** writes
   `InternalCurrentActorCellKeeper.Current = TestActor.Underlying` in four locations:
   - `EnsureImplicitSender` (line 58) — invoked on every `Host` / `Sys` access,
   - `StartActors` callback during host startup (line 122),
   - end of `InitializeAsyncCore` (line 197),
   - `BeforeTestStart` (line 225).

   Each assignment pins that test's `TestActor` onto the currently-executing ThreadPool thread.

3. **xUnit v3** defaults to parallel class execution (`parallelizeTestCollections: true`), where each test class is its own collection. Concurrent classes share the same ThreadPool, so pinned `Current` values get inherited by unrelated `await` continuations.

`ActorRefImplicitSenderExtensions.Tell(this IActorRef receiver, object message)` reads the pinned value via `ActorCell.GetCurrentSelfOrNoSender()`, which is how the foreign `TestActor` ends up on the wire.

### Related context in the Akka.NET codebase

Two pre-existing mechanisms are worth noting because they suggest the hosting TestKit diverged from, or bypassed, established Akka.NET patterns for this exact hazard:

- **`Akka.TestKit.Xunit` (core) already guards `EnsureImplicitSender` with `if (Current == null)`** (`akka.net/src/contrib/testkits/Akka.TestKit.Xunit/TestKit.cs:148`). `Akka.Hosting.TestKit` diverged from that discipline and overwrites unconditionally.
- **`Akka.NET` ships `ActorCellKeepingSynchronizationContext`** (`akka.net/src/core/Akka.TestKit/ActorCellKeepingSynchronizationContext.cs`) specifically to preserve `Current` across `await` continuations. `TestKitBase.InitializeTest` installs it — but `Akka.Hosting.TestKit.TestKit` overrides `InitializeTest` as a no-op (`TestKit.Shared.cs:202`) and runs the base init on the Akka *host startup* thread, not the test body thread. `SetSynchronizationContext` only affects the current thread, so the protective context is installed on the wrong thread and never takes effect inside `[Fact]` methods. A precedent regression of the same family (`Bugfix8144Spec`) is already documented in the Akka.NET repo for the xUnit v2 init-async case.

## Additional context

- Regression window: `Akka.Hosting.TestKit` ≥ 1.5.64 (xUnit v3 migration).
- Minimal repro files are attached to this issue: `Repro.csproj`, `Tests.cs`, `README.md`.
- Files referenced in this issue are at HEADs of `akkadotnet/akka.net` and `akkadotnet/Akka.Hosting` `dev` as of 2026-04-21.
- Related prior regression: `akka.net/src/contrib/testkits/Akka.TestKit.Xunit{,2}.Tests/Bugfix8144Spec.cs` documents the same class of `[ThreadStatic]` loss across `IAsyncLifetime.InitializeAsync`, fixed for xUnit v2 but not generalized to xUnit v3's parallel-class scheduling.

Aaronontheweb:
## Summary

Fixes #733 — `Akka.Hosting.TestKit` implicit-sender leaks across `ActorSystem`s under xUnit v3 parallel class execution.

### Problem

`InternalCurrentActorCellKeeper.Current` is `[ThreadStatic]` and does not flow through `ExecutionContext`. Under xUnit v3's default parallel-class scheduling (`MaxConcurrencySyncContext`), `await` continuations resume on ThreadPool threads where a sibling test pinned its cell, causing `Tell()` to use the wrong implicit sender. Replies then cross `ActorSystem` boundaries and land in the wrong `TestActor`'s mailbox.

### Root Cause of Prior Spike Hang

The akka.net PR #8174 approach (`AkkaCleanAmbientContextAttribute` + `ActorCellKeepingSynchronizationContext`) **replaces** xUnit's `SynchronizationContext` with one that dispatches via raw `ThreadPool.QueueUserWorkItem`. This works for `Akka.TestKit.Xunit` but hangs `Akka.Hosting.TestKit` because Hosting's `IHost` async startup lifecycle depends on xUnit's `MaxConcurrencySyncContext` scheduling.

### Fix

A **wrapping/decorating** `SynchronizationContext` that preserves xUnit's scheduler while pinning the ambient actor cell across `await` continuations:

- **`ActorCellKeepingSynchronizationContext`** — delegates `Post()`/`Send()` to the captured outer SC (preserving xUnit scheduling) while wrapping callbacks with a save/pin/restore window for `InternalCurrentActorCellKeeper.Current`. Falls back to `ThreadPool` dispatch when no outer SC exists.
- **`HostingCleanAmbientContextAttribute`** — `BeforeAfterTestAttribute` that captures the active SC before each test, installs the decorator, pins the cell, and restores everything in `After()`.
- **`EnsureImplicitSender()`** — guarded with `Current == null` to prevent overwriting the cell during actor message processing (matches akka.net's pattern).
- Removed two redundant cell-pinning writes now handled by the attribute.
- Enabled parallel test collections; 24 regression tests added (16 implicit-sender + 8 `INoImplicitSender`).

### Upstream Path

This proves the wrapping SC concept in Hosting. Once validated in CI, the approach can be upstreamed to akka.net PR #8174 to replace the raw-ThreadPool SC, then Hosting can switch to the upstream attribute.

## Changes

- `src/Akka.Hosting.TestKit/Internals/ActorCellKeepingSynchronizationContext.cs` — **New**: wrapping SC
- `src/Akka.Hosting.TestKit/Internals/HostingCleanAmbientContextAttribute.cs` — **New**: BeforeAfterTestAttribute
- `src/Akka.Hosting.TestKit/TestKit.cs` — Apply `[HostingCleanAmbientContext]`
- `src/Akka.Hosting.TestKit/TestKit.Shared.cs` — Guard `EnsureImplicitSender`, remove redundant writes
- `src/Akka.Hosting.TestKit.Tests/xunit.runner.json` — Enable parallel execution
- `src/Akka.Hosting.TestKit.Tests/ParallelAmbientContextSpec.cs` — **New**: 24 regression tests

## Test plan

- [x] 24 new `ParallelAmbientContextSpec` tests pass with parallel collections enabled (16 implicit-sender + 8 no-implicit-sender)
- [x] Full test suite: 303/303 pass, 0 failures, no hangs
- [x] Test discovery (`--list-tests`) completes without hanging
- [x] xUnit2 variant (`Akka.Hosting.TestKit.Xunit2`) builds clean, unaffected
- [ ] CI validation across all test projects


src/Akka.Hosting.TestKit/Internals/HostingCleanAmbientContextAttribute.cs
```diff
@@ -0,0 +1,61 @@
...
+using Xunit;
+using Xunit.v3;
+
+namespace Akka.Hosting.TestKit.Internals
+{
+    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false, Inherited = true)]
+    internal sealed class HostingCleanAmbientContextAttribute : BeforeAfterTestAttribute
+    {
+        [ThreadStatic]
+        private static SynchronizationContext? _previousContext;
```
Arkatufus: Is this advisable? `ThreadStatic` is the thing that causes the problem to begin with. Why is this code being replicated inside the attribute?

A static variable is a very bad code smell because this attribute is supposed to decouple the synchronization context from the thread runner, not coupling it even more.

Arkatufus (changes requested):
Big question

Arkatufus (approved):
LGTM

PR #743, opened 2026-05-08
fix: resolve GHSA-g94r-2vxg-569j by bumping OpenTelemetry to 1.10.0 and MEL floor to 9.0.0

Aaronontheweb:
## Summary

- Bumps `OpenTelemetry` minimum version floor in `Akka.Hosting.csproj` from `[1.9.0,)` to `[1.10.0,)`
- Bumps `OpenTelemetry.Extensions.Hosting` and `OpenTelemetry.Exporter.OpenTelemetryProtocol` in the example project from `1.9.0` to `1.10.0`

## Why

`OpenTelemetry.Api` 1.9.0 has a known moderate severity vulnerability ([GHSA-g94r-2vxg-569j](https://github.com/advisories/GHSA-g94r-2vxg-569j)). Because `Akka.Hosting` pinned `OpenTelemetry >= 1.9.0`, NuGet resolved to exactly 1.9.0, which pulled the vulnerable `OpenTelemetry.Api` 1.9.0 into downstream consumers.

This was discovered when a Dependabot PR in `akkadotnet/akka.net` ([#8192](https://github.com/akkadotnet/akka.net/pull/8192)) bumped `Akka.Management` to 1.5.67 (which depends on `Akka.Hosting` 1.5.67). The build failed with:

```
error NU1902: Warning As Error: Package 'OpenTelemetry.Api' 1.9.0 has a known moderate severity vulnerability
```

Bumping the minimum to `[1.10.0,)` resolves the transitive pull of the vulnerable package. `OpenTelemetry.Api` 1.10.0 is not affected by the vulnerability.

## Test plan

- [ ] CI passes (NuGet audit no longer flags vulnerable transitive dependency)

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -32,7 +32,7 @@
     <XunitRunneVisualstudio>3.1.5</XunitRunneVisualstudio>
     <Xunit3RunneVisualstudio>3.1.5</Xunit3RunneVisualstudio>
     <AkkaVersion>1.5.67</AkkaVersion>
-    <MicrosoftExtensionsVersion>[8.0.0,)</MicrosoftExtensionsVersion>
+    <MicrosoftExtensionsVersion>[9.0.0,)</MicrosoftExtensionsVersion>
     <SystemTextJsonVersion>[8.0.5,)</SystemTextJsonVersion>
   </PropertyGroup>
   <!-- SourceLink support for all Akka.NET projects -->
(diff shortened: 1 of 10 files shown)


Aaronontheweb:
Ran into a snag — `OpenTelemetry` 1.10.0 bumped its own `Microsoft.Extensions.*` dependencies to 9.0.0, which conflicts with Akka.Hosting's current `$(MicrosoftExtensionsVersion)` floor of `[8.0.0,)`. This causes NU1605 (package downgrade) errors throughout the solution.

The fix is larger than a simple version bump: we'd need to also raise `MicrosoftExtensionsVersion` to `[9.0.0,)`, which is a broader breaking change for downstream consumers still on .NET 8. Leaving this for further discussion on the right approach before proceeding.

PR #748, opened 2026-05-19
fix: widen Within test timeout to tolerate Windows CI scheduler jitter

Aaronontheweb:
## Root Cause

`Within_should_increase_max_timeout_by_the_provided_epsilon_value` uses `Within(1s, epsilon: 50ms)` with an implicit `ExpectNoMsg()`. On the new `windows-2025-vs2026` runner (GitHub is redirecting `windows-latest` by June 15, 2026), the OS scheduler fires the timer ~1.94s instead of ~1s, blowing the 1.05s ceiling.

## Fix

- Pass an explicit `ExpectNoMsg(1s)` so the wait duration is fixed and independent of the `Within` remaining time
- Widen the `Within` max to 3s to give the Windows scheduler room to breathe while still verifying the epsilon parameter works

## Test plan
- [ ] Windows CI passes `Within_should_increase_max_timeout_by_the_provided_epsilon_value`
- [ ] Linux CI continues to pass

### DIFF
--- a/src/Akka.Hosting.TestKit.Tests/TestKitBaseTests/WithinTests.cs
+++ b/src/Akka.Hosting.TestKit.Tests/TestKitBaseTests/WithinTests.cs
@@ -19,10 +19,12 @@ public class WithinTests : TestKit
     }
 
     [Fact]
-    public void Within_should_increase_max_timeout_by_the_provided_epsilon_value()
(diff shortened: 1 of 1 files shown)


Aaronontheweb:
Closing — changes already merged into dev via the 1.5.69 release.

